PR #24, opened 2020-04-14
Added 2p video transformation module

isaak-willett:
Added module containing the required 2p video data transformations.
These transformations were added to faciliate providing cut down data to
end user of ground truth app. The data transforms include, downsampling
of movie, box subset of movie, normalization of movie, and conversion of
movie to mp4 format.


segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+    Function to down-sample an array stored in h5 format to a desired FPS.
+    Follows the strategy specified in the function call, current strategies are
+    (random, maximum, average). Random takes a random sample of one frame from
+    window, maximum takes the maximum frame in window, average takes the mean
+    of window.
+    Args:
+        video_path: The path to the input video in str or Path format
+        input_fps: The FPS or Hz of the input video
+        output_fps: The desired output FPS or Hz
+        strategy: The down-sampling startegy to follow
```
djkapner, line 26: spelling

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+
+    Returns:
+        new_video: A new video down-sampled to desired FPS with selected
```
djkapner, line 29: this should document a type. Also the Args above.

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+        strategy
+    """
+    if os.path.exists(video_path):
```
djkapner, line 32: just curious, earlier you were a strong proponent of pathlib, I thought, and here you are using both `os` and `pathlib` when most things could be pathlib:
https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module
any reason?

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+        path_parts = str(video_path).split('.')
```
djkapner, line 33: `os.path.splitext()` or `PurePath.suffix`

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+        extension = path_parts[len(path_parts) - 1]
+        if extension == 'h5':
+            with h5py.File(video_path, 'r') as open_video:
+                video = open_video['data'][:]
+                return _down_sample_array(full_array=video,
```
djkapner, line 38: this does not need to be part of the context manager. You read the whole thing into memory with `['data'][:]`

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+                                          input_fps=input_fps,
+                                          output_fps=output_fps,
+                                          strategy=strategy)
+        else:
```
djkapner, line 42: maybe to-taste, but I don't really like burying function returns inside conditionals. Your conditionals could raise exceptions earlier and the function could end with a return.

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+            raise AttributeError('Video path provided not an h5 file, please'
+                                 'provide an h5 file')
+    else:
+        raise FileNotFoundError('Path specified doesnt exist or is not accessible')
+
+
+def _down_sample_array(full_array: np.ndarray, input_fps: int,
```
djkapner, line 49: `downsample` and `upsample` are commonly single words.

kschelonka, line 49: In addition we should add a `first` option, or something that approximates every `n` frames. Random is close but we want to have a little bit more uniformity. In addition, if we're having `random` we should also be able to control the seed in the function (otherwise we can't reproduce the results)

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+        output_fps: The desired output FPS or Hz
+        strategy: The down-sampling startegy to follow
+
+    Returns:
+        new_video: A new video down-sampled to desired FPS with selected
+        strategy
+    """
+    if output_fps > input_fps:
+        raise ValueError('Output FPS cannot be greater than input FPS')
+    sampling_ratio = int(input_fps / output_fps)
```
kschelonka, line 68: This is a straight truncation, so it approximates `floor`. Is this what we want?

djkapner, line 68: I think this will accumulate a time-dilation. For example 31Hz -> 4Hz, this will always be 8. The time axis of the downsampled movie won't be the time axis of the original movie. A way around this is to use `np.array_split` or `np.split` to make arrays of unequal size.

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+    return np.stack(new_frames)
+
+
+def get_transformed_center(coordinate_pair: Tuple[int, int],
+                           box_size: Tuple[int, int],
+                           video_shape: Tuple[int, int]):
+    """
+    Returns a coordinate pair that has been transformed to fit as closely to
+    original position as possible but still with box contained totally in video.
+    This is made to shift the coordinate if the subset box goes out of bound of
```
kschelonka, line 87: Good thinking about this edge case. Could you explain in the docstring what the "shift" is though?

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+        elif strategy == 'maximum':
+            new_frames.append(np.max(frames, axis=0))
+        elif strategy == 'average':
+            new_frames.append(np.mean(frames))
+    return np.stack(new_frames)
+
+
+def get_transformed_center(coordinate_pair: Tuple[int, int],
+                           box_size: Tuple[int, int],
+                           video_shape: Tuple[int, int]):
```
djkapner, line 83: don't you typically put a type hint for the return?

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+        if (coordinate_pair[0] + (math.ceil(box_size[0]/2))) > video_shape[0] - 1:
+            transformed_x = video_shape[0] - math.ceil(box_size[0] / 2) - 1
+
+        # can't be centered down
+        if (coordinate_pair[1] + (math.ceil(box_size[1]/2))) > video_shape[1] - 1:
+            transformed_y = video_shape[1] - math.ceil(box_size[1] / 2) - 1
+
+        return transformed_x, transformed_y
+
+    else:
```
djkapner, line 120: again, could raise this exception at the start, and then the return is not inside a conditional and the whole function isn't indented.

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+    transformed_video = np.stack(transformed_stack)
+    return transformed_video
+
+
+def generate_max_ave_proj_image(video: np.ndarray, projection_type: str= 'average'):
+    """
+    Returns a maximum projection or an average projection of a video in
+    numpy array format
+    Args:
+        video: The video to generate the projection
```
djkapner, line 160: I think the type and dimension shape needs to be called out

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+    Returns:
+
+    """
+    transformed_coordinates = get_transformed_center(coordinate_pair, box_size,
+                                                     video_array.shape)
+    left_column = transformed_coordinates[0] - math.ceil(box_size[0] / 2)
+    right_column = transformed_coordinates[0] + math.ceil(box_size[0] / 2)
+    up_row = transformed_coordinates[1] - math.ceil(box_size[1] / 2)
+    down_row = transformed_coordinates[1] + math.ceil(box_size[1] / 2)
+    transformed_stack = []
```
kschelonka, line 147: You're reallocating the memory for this large array 2x -- in this array and then converting it to a numpy array with `stack`. I suggest either allocate a numpy array from the beginning or operate directly on the original array.

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+
+        return transformed_x, transformed_y
+
+    else:
+        raise ValueError('Box is larger or equal in size in one or more'
+                         'dimensions to original video shape. Provide a '
+                         'box size smaller than original video.')
+
+
+def get_centered_coordinate_box(coordinate_pair: Tuple[int, int],
```
kschelonka, line 126: I may ask you to rework this a bit if the video transformations take too long, but performance isn't our number one concern since this isn't going to be used very often.

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+def normalize_video(video: np.ndarray):
+    """
+    Function to normalize video by its global max and maximum 8 bit value
+    Args:
+        video: Video to be normalized
+    Returns:
+        norm_frames: normalized video
+    """
+    global_max = video.max()
+    norm_frames = []
```
kschelonka, line 181: Similar to above comment about allocation. For this one just create an empty array and insert the appropriate frames.

segmentation_labeling_app/transforms/transformations.py
djkapner: yes, I had some comment on this farther down

segmentation_labeling_app/transforms/transformations.py
kschelonka, line 181: Here you can just create a new array directly by applying a function, so you don't need a loop or copy.

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+        return np.mean(video)
+    elif projection_type == 'maximum':
+        return np.max(video, axis=0)
+
+
+def normalize_video(video: np.ndarray):
+    """
+    Function to normalize video by its global max and maximum 8 bit value
+    Args:
+        video: Video to be normalized
```
kschelonka, line 176: Clarify dimensions here please

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+                                box_size: Tuple[int, int],
+                                video_array: np.ndarray):
+    """
+    Function to get video subset centered around coordinate pair with
+    size specified by box size tuple. Function takes a subset of each frame
+    and stacks together to get final video.
+    Args:
+        coordinate_pair: the coordinate on which to center the subset
+        box_size: the size of the subset box
+        video_array: the video to take the subset from
```
kschelonka, line 136: Clarify dimensions here please

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+                           video_shape: Tuple[int, int]):
+    """
+    Returns a coordinate pair that has been transformed to fit as closely to
+    original position as possible but still with box contained totally in video.
+    This is made to shift the coordinate if the subset box goes out of bound of
+    the video shape.
+    Args:
+        coordinate_pair: the original coordinate location for centering
+        box_size: the size of the box for subset
+        video_shape: the shape of the original video
```
kschelonka, line 92: Clarify here that this is the x-y dimensions, not the true `np.shape` of the video (which would include the time axis as 0 in our data)

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+    norm_frames = []
+    for frame in video:
+        norm_frame = copy.deepcopy(frame)
+        norm_frame = np.uint8(norm_frame / global_max * 255)
+        norm_frames.append(norm_frame)
+    norm_frames = np.stack(norm_frames)
+    return norm_frames
+
+
+def transform_to_mp4(video: np.ndarray, output_path: str):
```
kschelonka, line 190: Similar to the comment in the centered-box extraction, if this transformation takes too long in practice we may need to rework this approach.

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    writer = mpg.write_frames(output_path,
+                              video[0].shape,
+                              pix_fmt_in="gray",
+                              pix_fmt_out="gray",
+                              fps=30)
+    writer.send(None)
+    norm_video = normalize_video(video)
+    for frame in video:
+        frame = np.uint8(frame / frame.max() * 255)
```
kschelonka, line 209: Shouldn't we use the global max here?

tests/test_transformations.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+@pytest.fixture()
+def video_fixture():
+    current_path = (Path(__file__)).parent
+    video_path = current_path / 'resources' / 'test_video.h5'
+    yield video_path
+
+
+def test_video_random_downsampling(video_fixture):
```
kschelonka, line 22: Good coverage of the different types. I think it would be a lot less verbose if you parametrized the tests and included the downsampling type argument in the parametrization.

tests/test_transformations.py
```diff
@@ -0,0 +1,177 @@
...
+                                                    output_fps=output_fps)
+
+    with h5py.File(video_fixture, 'r') as open_video:
+        video = open_video['data'][:]
+        match_count = 0
+        new_video_index = 0
+        sampling_ratio = int(input_fps / output_fps)
+        expected_count = math.ceil(len(video) / int((input_fps / output_fps)))
+
+        for i in range(0, len(video), sampling_ratio):
```
kschelonka, line 36: This applies to all similar tests -- you should be able to generate exactly the data you expect, and then use numpy.testing module to assert array equality.

tests/test_transformations.py
```diff
@@ -0,0 +1,177 @@
...
+        for frame in reader:
+            gray_frame = cv2.cvtColor(frame, cv2.COLOR_RGB2GRAY)
+            frames.append(gray_frame)
+        frames = np.stack(frames)
+
+        # codec for mp4 encoding changes the video slightly between h5 and mp4
+        # and vice versa, so numbers are not exactly the same. Instead general
+        # structure of video array loaded. Cv color conversion does not rectify
+        # this error unfortunately, this has been attempted and small errors
+        # persist
```
kschelonka, line 166: How small? Small enough for `numpy.testing.assert_close` etc?

tests/test_transformations.py
```diff
@@ -0,0 +1,177 @@
...
+import h5py
+import numpy as np
+import cv2
+import imageio
+
+import segmentation_labeling_app.transforms.transformations as transformations
+
+
+@pytest.fixture()
+def video_fixture():
```
kschelonka, line 16: We don't need to use a real video file. This one is quite big for a test resource, and we don't know exactly what's in it so it's hard to make the `expected` result. Just generate a small sample 3d numpy array such that it's easy to test (e.g. binary values, `n` * 2 * 2 dimensions.

tests/test_transformations.py
```diff
@@ -0,0 +1,177 @@
...
+    assert transformed_center == expected_pair
+
+
+@pytest.mark.parametrize(("coordinate_pair, window_size"),
+                         [((10, 10), (6, 6)),
+                          ((10, 10), (12, 12)),
+                          ((510, 510), (10, 10)),
+                          ((10, 510), (10, 10)),
+                          ((510, 10), (10, 10))])
+def test_video_subset_coord_pair(coordinate_pair, window_size):
```
kschelonka, line 115: Testing this would be a lot easier if the coordinate extraction was separated from the video writing in the function, don't you think?

segmentation_labeling_app/transforms/transformations.py
isaak-willett: good catch

segmentation_labeling_app/transforms/transformations.py
isaak-willett: if it's documented in the function signature the type isn't to be specified in pep I believe

segmentation_labeling_app/transforms/transformations.py
isaak-willett: It was to be agnostic to str vs path input, you can use os on path objects but not the other way around. I think for simplicity I'll remove the agnostic input type.

segmentation_labeling_app/transforms/transformations.py
isaak-willett: changed

segmentation_labeling_app/transforms/transformations.py
isaak-willett: I was originally going to load all the video into memory since we'll be on the cluster but I won't do that.

segmentation_labeling_app/transforms/transformations.py
isaak-willett: I agree, changed the structure, now raises exceptions first then proceeds if not exceptions

segmentation_labeling_app/transforms/transformations.py
isaak-willett: Thanks for the tip

isaak-willett: Added random seeding as well as first and last options of sampling. Good point!

segmentation_labeling_app/transforms/transformations.py
isaak-willett: Fixed time dilation I think, I'll need some confirmation this is fixed when I update it.

segmentation_labeling_app/transforms/transformations.py
isaak-willett: Yes, good catch

segmentation_labeling_app/transforms/transformations.py
isaak-willett: Better explanation and example included in docstring

segmentation_labeling_app/transforms/transformations.py
isaak-willett: good call

segmentation_labeling_app/transforms/transformations.py
isaak-willett: Good idea

segmentation_labeling_app/transforms/transformations.py
isaak-willett: Good call, I allocated up front and then altered memory as the function loops.

segmentation_labeling_app/transforms/transformations.py
isaak-willett: No we don't need this it's an artifact from the previous testing I was doing, good catch.

segmentation_labeling_app/transforms/transformations.py
isaak-willett: Good catch

tests/test_transformations.py
isaak-willett: Sounds good, moved to a in memory movie. I'll remove the file from the commit histroy.

tests/test_transformations.py
isaak-willett: I agree that it would be easier to read if parametrized, the seed is now controlled so that won't be an issue

tests/test_transformations.py
isaak-willett: maybe...

isaak-willett: I don't think so, it comes from transitioning from grey to colored to back since mp4 requires rgb writing and reads out as rgb as well.

tests/test_transformations.py
isaak-willett: I think the math is trivial enough its not too cumbersome, it would only be four lines to get the column and row information.

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,225 @@
...
+
+    Returns:
+
+    """
+    norm_video = normalize_video(video)
+    writer = mpg.write_frames(output_path,
+                              video[0].shape,
+                              pix_fmt_in="gray",
+                              pix_fmt_out="gray",
+                              fps=30)
```
kschelonka, line 220: This should be based on the downsampling rate. If we downsample to 4Hz/fps) and then set it to 30 here it will be quite fast. Maybe then just add it as an arg in your function?

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,225 @@
...
+            transformed_x = video_shape[0] - math.ceil(box_size[0] / 2) - 1
+
+        # can't be centered down
+        if (coordinate_pair[1] + (math.ceil(box_size[1]/2))) > video_shape[1] - 1:
+            transformed_y = video_shape[1] - math.ceil(box_size[1] / 2) - 1
+
+    return transformed_x, transformed_y
+
+
+def get_centered_coordinate_box_video(coordinate_pair: Tuple[int, int],
```
kschelonka, line 147: Add test coverage here please

segmentation_labeling_app/transforms/transformations.py
isaak-willett: That's my mistake I thought I had it as a variable.

segmentation_labeling_app/transforms/transformations.py
isaak-willett: test_video_subset_coord_pair is the test coverage for this function

kschelonka: That test doesn't have adequate coverage for this function. I will show you an example.

kschelonka: I updated this branch with a test that has two cases (note: does not cover all edge cases). I strongly encourage test-driven development. These cases will get you started in debugging this function, but you'll need to add more.

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -0,0 +1,229 @@
...
+                                      box_size: Tuple[int, int],
+                                      video_array: np.ndarray):
+    """
+    Function to get video subset centered around coordinate pair with
+    size specified by box size tuple. Function takes a subset of each frame
+    and stacks together to get final video.
```
djkapner, line 153: I think your docstring needs to specify row/col or x/y for the inputs. Because `coordinate_pair` and `box_size` are reversed from each other.

segmentation_labeling_app/transforms/transformations.py
djkapner: it is still hard-coded to 30. You want `fps=fps`

isaak-willett: thanks for finding that, I feel really dumb leaving that in hah.

tests/test_transformations.py
```diff
@@ -0,0 +1,127 @@
...
+    # structure of video array loaded. Cv color conversion does not rectify
+    # this error unfortunately, this has been attempted and small errors
+    # persist
+
+    assert frames.shape == video_fixture.shape
+
+    os.remove(output_path)
+
+
+@pytest.mark.parametrize("video,coord,box_size,expected",
```
kschelonka, line 101: It's helpful to explain your test cases in the comments when it's not immediately clear.

segmentation_labeling_app/transforms/transformations.py
kschelonka: To add to Dan -- we should follow standard convention, same as numpy (row x col / H x W)

isaak-willett: Went and corrected all of the instances of this.

tests/test_transformations.py
```diff
@@ -0,0 +1,127 @@
...
+def test_coordinate_window_shift(coordinate_pair, window_size, video_shape,
+                                 expected_pair):
+    transformed_center = transformations.get_transformed_center(coordinate_pair,
+                                                                     window_size,
+                                                                     video_shape)
+    assert transformed_center == expected_pair
+
+
+def test_mp4_conversion(video_fixture):
+    output_path = Path(__file__).parent
```
kschelonka, line 73: Use tempdir please.

tests/test_transformations.py
```diff
@@ -0,0 +1,127 @@
...
+     (test_video, (0, 10), (3, 3), np.stack([np.array([[10, 11, 12], [15, 16, 17],
+                                                       [20, 21, 22]])] * 2)),
+     (test_video, (10, 0), (3, 3), np.stack([np.array([[2, 3, 4], [7, 8, 9],
+                                                       [12, 13, 14]])] * 2)),
+     (test_video, (2, -1), (3, 3), np.stack([np.array([[1, 2, 3], [6, 7, 8],
+                                                        [11, 12, 13]])] * 2)),
+     (test_video, (-1, 2), (3, 3), np.stack([np.array([[5, 6, 7], [10, 11, 12],
+                                                       [15, 16, 17]])]*2)),
+     (test_video, (0, 0), (5, 5), test_video),
+     (test_video, (2, 2), (0, 0), np.zeros(shape=(2, 0, 0)))]
```
kschelonka, line 122: What if the box is too big for the video?

tests/test_transformations.py
```diff
@@ -0,0 +1,127 @@
...
+        [[[5, 5], [5, 5]], [[2, 2], [2, 2]],
+         [[4, 4], [4, 4]], [[0, 0], [0, 0]],
+         [[3, 2], [10, 7]], [[11, 1], [0, 5]],
+         [[8, 2], [12, 1]], [[6, 3], [2, 8]],
+         [[11, 3], [1, 17]], [[8, 3], [9, 0]]]
+    )
+    yield random_video
+
+
+@pytest.mark.parametrize(("input_fps, output_fps, random_seed, strategy, "
```
kschelonka, line 30: Please add cases for when input and output fps are not perfectly divisible

tests/test_transformations.py
isaak-willett: Sounds good

tests/test_transformations.py
isaak-willett: Raises an exception in the function for centered coordiantes, I'm not sure how to test for a raised exception.

tests/test_transformations.py
isaak-willett: done

tests/test_transformations.py
isaak-willett: done

tests/test_transformations.py
djkapner: Here's an example:
https://github.com/AllenInstitute/segmentation-labeling-app/blob/8abfdb22845bca4fd01c28255d6c115d16172a8b/tests/test_transfer_utils.py#L83-L84
and the docs:
https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

tests/test_transformations.py
djkapner: looks like this comment is outdated, but, `numpy.allclose` and `numpy.testing.assert_allclose` have parameters for adjusting required relative and absolute tolerances for cases where default machine precision is too stringent.
https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html

tests/test_transformations.py
isaak-willett: This doesn't explain how to use the pytest raise for only one parameterized case

kschelonka: Also see https://github.com/AllenInstitute/segmentation-labeling-app/blob/master/tests/transforms/test_array_utils.py

Line 52 I have the separated test with a clear name to say its purpose.

isaak-willett:
The following script has been run to produce the below output. A video conversion has also been run and evaluated for accuracy. Can be provided if curious.
```
from pathlib import Path

import h5py

import segmentation_labeling_app.transforms.transformations as transformations

movie_path = Path('/allen/aibs/informatics/isaak/suite2p_demixing/movies/concat_31Hz_0.h5')

downsampled_movie = movie_path.parent / 'downsample_movie.h5'

with h5py.File(movie_path, 'r') as open_movie:
    movie = open_movie['data']
    with h5py.File(downsampled_movie) as open_downsample:
        down_movie = open_downsample['data']
        sampling_ratio = movie.shape[0] / down_movie.shape[0]
        print('Original Shape: ', movie.shape)
        print('Downsampled Shape: ', down_movie.shape)
        print('Sampling Ration: ', sampling_ratio)
```

```
Original Shape:  (115766, 512, 512)
Downsampled Shape:  (9647, 512, 512)
Sampling Ration:  12.000207318337306
```

PR #25, opened 2020-04-14
Added ROIs class and util modules

isaak-willett:
Added ROIs class to hold required loading and morphological functions.
This was added to correctly identify the edge pixels of the inner
outline of an ROI in a 2D image. This was undertaken by identifying the
edges and the center then moving in along line by stroke size provided.
Query utils were added to load ROI data straight from LIMs db.


segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,215 @@
...
+    def __init__(self,
+                 coo_rows: Union[np.array, List[int]],
+                 coo_cols: Union[np.array, List[int]],
+                 coo_data: Union[np.array, List[float]],
+                 image_shape: Tuple[int, int],
+                 segmentation_id: int,
+                 roi_id: int):
+        self.coo_rows = coo_rows
+        self.coo_cols = coo_cols
+        self.coo_data = coo_data
```
djkapner, line 31: your `coo_*` properties are going to be duplicated between `ROI` and `ROI._sparse_coo`. Maybe not in space, if they are by reference, but, it might be more clear if the matrix is only described in one place.

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,215 @@
...
+                   coo_data=roi['coo_data'][0],
+                   image_shape=shape,
+                   segmentation_id=segmentation_id,
+                   roi_id=roi_id)
+
+    @property
+    def sparse_coo(self):
+        """
+        Produces a sparse coo matrix scipy object from the
+        Returns: A sparse_csc scipy object for the data
```
djkapner, line 80: csc, coo, could be one line says the same  thing

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,215 @@
...
+            port=label_vars.port,
+            password=label_vars.password)
+        return ROI(coo_rows=roi['coo_rows'][0],
+                   coo_cols=roi['coo_cols'][0],
+                   coo_data=roi['coo_data'][0],
+                   image_shape=shape,
+                   segmentation_id=segmentation_id,
+                   roi_id=roi_id)
+
+    @property
```
djkapner, line 76: unclear to me why this is a property, and if it is,why is there not a setter()? I'm not sure why one would update this,but to do so, one would have to update self.coo_* , set self._sparse_coo to None and then call the property. With a setter and a getter, it all would work out more cleanly. I could update by roi.set_sparse(vals) or something.
Anyway, I'm not sure why the coo is not just an attribute made on class instantiation. What's the use-case for it being mutable?

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,215 @@
...
+        """
+        if self._sparse_coo is None:
+            self._sparse_coo = coo_matrix((
+                self.coo_data,
+                (self.coo_rows,
+                 self.coo_cols)),
+                shape=self.image_shape)
+        return self._sparse_coo
+
+    @property
```
djkapner, line 90: Do not need this? If the user wants this, they can just `roi._sparse_coo.toarray()`

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,215 @@
...
+
+        Returns:
+
+        Notes:
+            This function will likely become deprecated as we work on more
+            applicable binary thresh holding
+        """
+        return np.where(self.dense_matrix >= threshold, 1, 0)
+
+    def generate_roi_inner_edge_coordinates(self, stroke_size: int,
```
djkapner, line 114: I think you can do a lot of this with opencv. Did you check out `contours`?
https://docs.opencv.org/3.4/d4/d73/tutorial_py_contours_begin.html

Also, I noticed opencv was a requirement, but here you are using ndimage heavily. Using multiple image-processing libraries tends to lead to a lot of conversions and confusion. If you can simply base on ndimage OR opencv, it ends up being easier in the end.

segmentation_labeling_app/utils/math_utils.py
```diff
@@ -0,0 +1,18 @@
+from typing import Tuple
+import math
+
+
+def get_magnitude(coord_one: Tuple[int, int],
```
djkapner, line 5: this is a whole file for `np.linalg.norm()`

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        Returns a list of coordinates in a video frame for the inner outline
+        of an ROI given a desired stroke size. Looks at each border pixel and
+        moves away from edge by stroke specified pixels. This process is completed
+        by defining the center of the roi and creating a unit vector from the
+        center to the edge point. This unit vector is then multiplied by the stroke
+        and a slight weight to correct for non int centers. The movement vector
+        is added to the edge point and the new point is then set to true in
+        the return mask.
+        Args:
```
djkapner, line 95: This is a pretty verbose docstring, and after reading it, I'm still not sure what an `inner outline` is, nor a `stroke size`.
Are you just trying to find the edge of an ROI?
I see you've used contours in a call from here, to respond to my earlier comment.
Can you explain in words what you're trying to do here. I suspect there is some canned routine to just do it.

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,170 @@
...
+            that has been slightly eroded
+            coordinate_list: A list of coordinates for the edge of the ROI
+            pushed in by stroke pixel count
+        """
+        edge_coordinates = self.get_edge_points(threshold=threshold)
+        center = self.get_centroid_of_thresholded_mask(threshold=threshold)
+        inner_edge_mask = np.zeros(shape=self.image_shape)
+        for edge_coordinate in edge_coordinates[0]:
+            # get distance to center
+            vector = np.array([(center[0] - edge_coordinate[0][0]),
```
djkapner, line 111: I think this whole function can probably be out-sourced to some package, but, fwiw, you should check out:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.cdist.html
this would probably help you vectorize this loop, making it faster and easier to read.

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,170 @@
...
+                we multiply it by a defined weight to overweight the movement
+                just slightly to compensate for imperfect stroke movement.
+                """
+                movement_vector = stroke_weight * stroke_size * unit_vector
+                new_position_x = int(round(edge_coordinate[0][0] + movement_vector[0]))
+                new_position_y = int(round(edge_coordinate[0][1] + movement_vector[1]))
+                new_position_vec = np.array([new_position_x, new_position_y])
+                inner_edge_mask[new_position_vec[0]][new_position_vec[1]] = 1
+            else:
+                raise ValueError("Stroke size too large, cannot get inner edge "
```
djkapner, line 129: Is this because you are hitting the edge of the frame?

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,170 @@
...
+            port=label_vars.port,
+            password=label_vars.password)
+        return ROI(coo_rows=roi['coo_rows'][0],
+                   coo_cols=roi['coo_cols'][0],
+                   coo_data=roi['coo_data'][0],
+                   image_shape=shape,
+                   segmentation_id=segmentation_id,
+                   roi_id=roi_id)
+
+    def generate_binary_mask_from_threshold(self, threshold: float):
```
djkapner, line 70: I think most of your class methods would be much more transparent and more easily-tested as stand-alone functions where the inputs and outputs can be clearly stated.

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,170 @@
...
+
+class ROI:
+    """
+    This module is used for loading the ROIs from LIMs DB tables and
+    contains pre processing methods for ROIs loaded from LIMs. These methods
+    are used to define drawing parameters as well as other ROI class
+    methods useful for post processing required to display to end user.
+    """
+
+    def __init__(self,
```
djkapner, line 20: I think this needs a docstring.

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,170 @@
...
+        self._sparse_coo = coo_matrix((coo_data, (coo_rows, coo_cols)),
+                                      shape=image_shape)
+
+    @classmethod
+    def roi_from_query(cls, segmentation_id:int,
+                       roi_id: int) -> "ROI":
+        """
+        Queries and builds ROI object by querying LIMS table for
+        produced labeling ROIs.
+        Args:
```
djkapner, line 39: what format is this docstring? Why aren't there types and dimensions?

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,170 @@
...
+                   segmentation_id=segmentation_id,
+                   roi_id=roi_id)
+
+    def generate_binary_mask_from_threshold(self, threshold: float):
+        """
+        Simple binary mask from a provided threshold.
+        Args:
+            threshold: The threshold to compare values
+
+        Returns:
```
djkapner, line 76: returns?

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,230 @@
...
+        return mask_matrix
+
+    def create_manifest_json(self, source_ref: str,
+                             video_source_ref: str, max_source_ref: str,
+                             avg_source_ref: str, trace_source_ref: str,
+                             roi_data_source_ref: str) -> str:
+        """
+        Function to make the manifest json string required for sagemaker
+        ground truth to read the data in successfully. The strings are
+        S3 bucket URIs for each of the required data.
```
djkapner, line 189: keys should be local file paths not S3 bucket URIs

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,230 @@
...
+        Args:
+            source_ref: Location of ROI mask png
+            video_source_ref: Location of 2P video for the ROI
+            max_source_ref: Location of the maximum projection for the ROI
+            avg_source_ref: Location of the average projection for the ROI
+            trace_source_ref: Location of the trace data for the ROI
+            roi_data_source_ref: Location of the ROI coordinate data
+
+        Returns:
+            dictionary: returns dictionary as binary json string
```
djkapner, line 199: not sure about `binary`? Docs for `json.dumps` say `Serialize obj to a JSON formatted str`

segmentation_labeling_app/utils/query_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    Args:
+        db_file: database file
+
+    Returns:
+
+    """
+    conn = None
+    try:
+        conn = sqlite3.connect(db_file)
+    except Error as err:
```
djkapner, line 50: I'm not sure what you're attempting to catch here, but your try except is exactly the same as :
```conn = sqlite3.connect(db_file)```
It's going to except and print the exception? am i missing something?

File: segmentation_labeling_app/utils/query_utils.py [hunk 17]
```diff
@@ -0,0 +1,113 @@
...
+    """
+    Creates a table from the create_table_sql statement
+    Args:
+        conn: Connection object
+        create_table_sql: a CREATE TABLE statement
+
+    Returns:
+
+    """
+    try:
```
djkapner, line 65: again, you are catching a sqlite3.Error and printing it, but that would happen even without your try/except

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,230 @@
...
+        return json.dumps(dictionary)
+
+    def write_roi_to_db(self, database_file_path: Path,
+                        transform_hash: str,
+                        ophys_segmentation_commit_hash: str,
+                        creation_date: str,
+                        manifest_json: str,
+                        upload_date: str = None):
+        # connect to the database file
+        db_connection = query_utils.create_connection_sqlite(database_file_path.as_posix())
```
djkapner, line 218: I don't think your function `create_connection_sqlite` adds any functionality on top of `sqlite3.connect`

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,230 @@
...
+
+        # create table, sql query handles the checking if already exists
+        query_utils.create_table_sqlite(db_connection, sql_create_rois_table)
```
djkapner, line 221: I don't think your function `create_table_sqlite` adds any functionality that `conn.execute()` would do

segmentation_labeling_app/utils/query_utils.py
```diff
@@ -0,0 +1,113 @@
...
+        c = conn.cursor()
+        c.execute(create_table_sql)
+    except Error as err:
+        print(err)
+
+
+def insert_into_sqlite_table(conn: sqlite3.Connection, sqlite_insert_command: str,
```
djkapner, line 72: again, not seeing the value in wrapping execute with a function that calls execute?

tests/test_rois.py
```diff
@@ -0,0 +1,225 @@
...
+     [0.75, 0.8, 0.9, 0.85], (3, 3), 1, 1, 'test_hash', 'test_hash',
+     'test_date', 'test_date', json.dumps('test_manifest'))
+])
+def test_roi_db_table_write(coo_rows, coo_cols, coo_data, video_shape,
+                            segmentation_id, roi_id, transform_hash,
+                            ophys_segmentation_commit_hash, creation_date,
+                            upload_date, manifest_json):
+    test_roi = roi_module.ROI(coo_rows, coo_cols, coo_data,
+                              video_shape, segmentation_id, roi_id)
+    database_file = Path(__file__).parent / 'test_db.db'
```
djkapner, line 200: you should put this test file in a `tmpdir` somewhere using pytest.tmpdir

tests/test_rois.py
```diff
@@ -0,0 +1,225 @@
...
+    assert db_roi[2] == ophys_segmentation_commit_hash
+    assert db_roi[3] == creation_date
+    assert db_roi[4] == upload_date
+    assert db_roi[5] == manifest_json
+    assert db_roi[6] == test_roi.experiment_id
+    assert db_roi[7] == test_roi.roi_id
+
+    db_connection.close()
+
+    os.remove(database_file)
```
djkapner, line 225: and then you won't need this

isaak-willett:
The following was run to validate that the ROIs can be loaded and used successfully using the query constructor.
```
import psycopg2
import pandas as pd

import segmentation_labeling_app.rois.rois as roi_module
import segmentation_labeling_app.utils.query_utils as query_utils

label_vars = query_utils.get_labeling_env_vars()

test_roi = roi_module.ROI.roi_from_query(experiment_id=1,
                                         roi_id=15)
print('mask: ', test_roi._sparse_coo.toarray())
binary_mask = test_roi.generate_binary_mask_from_threshold(threshold=0.5)
print('binary_mask: ', binary_mask)
edge_points = test_roi.get_edge_points(threshold=0.5)
print('edge points: ', edge_points)
edge_mask = test_roi.get_edge_mask(threshold=0.5)
print('edge mask', edge_mask)
dilated_contours = test_roi.create_dilated_contour_mask(stroke_size=3, threshold=0.5)
print('dilated contours: ', dilated_contours)
```

The script ran without error and produced the following output
```
mask:  [[0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]
 ...
 [0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]]
binary_mask:  [[0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 ...
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]]
edge points:  [[326 299]
 [325 300]
 [324 300]
 [323 300]
 [322 300]
 [323 300]
 [324 300]
 [325 300]
 [326 299]
 [327 299]
 [328 299]
 [329 299]
 [330 299]
 [331 299]
 [330 299]
 [329 299]
 [328 299]
 [327 299]]
edge mask [[0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]
 ...
 [0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]]
dilated contours:  [[0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]
 ...
 [0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]]```
Showing all the functionality runs as expected.

File: tests/test_rois.py [hunk 23]
```diff
@@ -0,0 +1,220 @@
...
+])
+def test_binary_mask(coo_rows, coo_cols, coo_data, video_shape,
+                     segmentation_id, roi_id, threshold, expected_mask):
+    test_roi = roi_module.ROI(coo_rows, coo_cols, coo_data,
+                              video_shape, segmentation_id, roi_id)
+    generated_mask = test_roi.generate_binary_mask_from_threshold(threshold=threshold)
+    assert np.array_equal(generated_mask, expected_mask)
+
+
+@pytest.mark.parametrize(("coo_rows, coo_cols, coo_data, video_shape,"
```
kschelonka, line 27: Please add following cases:
* Single pixel mask
* Empty mask
* Masks with non-contiguous points
* Some values below threshold
* All values below threshold

isaak-willett:
binary_mask
dilated_mask
new_dilation

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,240 @@
...
+            user=label_vars.user,
+            host=label_vars.host,
+            database=label_vars.database,
+            port=label_vars.port,
+            password=label_vars.password)
+        return ROI(coo_rows=roi[0]['coo_row'],
+                   coo_cols=roi[0]['coo_col'],
+                   coo_data=roi[0]['coo_data'],
+                   image_shape=shape,
+                   experiment_id=experiment_id,
```
djkapner, line 92: I think we're going to want to propogate the `ophys_experiment_id` through the pipeline. Here, you're propogating the `segmentation_run_id` which are different things. Have a look at the table designs again:
https://github.com/AllenInstitute/informatics_design_docs/tree/master/ophys/segmentation#table-designs

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,240 @@
...
+        curr = db_connection.cursor()
+        curr.execute(sql_create_rois_table)
+
+        roi_task = (transform_hash, ophys_segmentation_commit_hash,
+                    creation_date, upload_date, manifest,
+                    self.experiment_id, self.roi_id)
+        # add the roi to the table
+        curr.execute(sql_insert_roi, roi_task)
+        unique_id = curr.lastrowid
+        db_connection.commit()
```
djkapner, line 238: It is probably a good habit to include `db_connection.close()`.I suspect for sqlite, this is not too much of an issue, but I've run into issues in other systems at scale when forgetting these closing calls.

tests/test_rois.py
```diff
@@ -0,0 +1,247 @@
...
+                          "segmentation_id, roi_id, threshold, expected_edges"),
```
kschelonka, line 28: Thanks for adding test cases. Please make notes in docstring or comments next to the cases which ones they represent so that it's easier for the reader.

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,241 @@
...
+            optional_array: an optional array to overload the function and
+            create edge points from different mask
+        Returns:
+            Contours: a list of coordinates that contain edge points
+        """
+        if optional_array is None:
+            optional_array = self.generate_binary_mask_from_threshold(threshold)
+        normalized_array = np.uint8(optional_array * 255)
+        contours, _ = cv2.findContours(normalized_array,
+                                       cv2.RETR_TREE,
```
kschelonka, line 159: Use CV_RETR_LIST over the tree since you're not using the hierarchical information

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,241 @@
...
+            dilated_mask = self.get_edge_mask(threshold=1,
+                                              optional_array=dilated_mask)
+            final_mask = final_mask + dilated_mask
+        final_mask = np.where(final_mask >= 1, 1, 0)
+        return final_mask
+
+    def get_edge_points(self, threshold: float,
+                        optional_array: np.array = None):
+        """
+        Returns a list of edge points. Computes edges using binary erosion and
```
kschelonka, line 145: Please update docstring

segmentation_labeling_app/rois/rois.py
```diff
@@ -0,0 +1,241 @@
...
+            threshold: a float to threshold all values against to create binary
+                       mask
+            optional_array: an optional array to overload the function and
+            create edge points from different mask
+        Returns:
+            Contours: a list of coordinates that contain edge points
+        """
+        if optional_array is None:
+            optional_array = self.generate_binary_mask_from_threshold(threshold)
+        normalized_array = np.uint8(optional_array * 255)
```
kschelonka, line 157: I don't think you need to do this step: [From the documentation](https://docs.opencv.org/2.4/modules/imgproc/doc/structural_analysis_and_shape_descriptors.html)

> "Non-zero pixels are treated as 1’s. Zero pixels remain 0’s, so the image is treated as binary ."

djkapner: more recent doc:
https://docs.opencv.org/3.4/d3/dc0/group__imgproc__shape.html#ga556a180f43cab22649c23ada36a8a139

tests/test_rois.py
kschelonka: Please also update these tests once the coordinate values are returned in the correct order (row,col)

segmentation_labeling_app/rois/rois.py
kschelonka: Thanks Dan! /latest/ wasn't working.

segmentation_labeling_app/rois/rois.py
```diff
@@ -142,27 +142,27 @@ def create_dilated_contour_mask(self, stroke_size: int,
...
-        normalized_array = np.uint8(optional_array * 255)
-        contours, _ = cv2.findContours(normalized_array,
-                                       cv2.RETR_TREE,
+        contours, _ = cv2.findContours(optional_array,
+                                       cv2.RETR_LIST,
                                        cv2.CHAIN_APPROX_NONE)
         if len(contours) > 0:
             return np.unique(np.squeeze(np.concatenate(contours), axis=1), axis=0)
         else:
             print('no edges detected')
```
djkapner, line 164: we should not be printing, we should be [logging](https://docs.python.org/3.8/library/logging.html).

segmentation_labeling_app/rois/rois.py
```diff
@@ -142,27 +142,27 @@ def create_dilated_contour_mask(self, stroke_size: int,
...
-            return None
+            return []
```
djkapner, line 165: your docstring says returns a numpy array, but in this case, you're returning an empty list. Probably wherever this came up, you anticipated this usage, but not great for future maintenance/use.

segmentation_labeling_app/rois/rois.py
```diff
@@ -142,27 +142,27 @@ def create_dilated_contour_mask(self, stroke_size: int,
     def get_edge_points(self, threshold: float,
```
djkapner, line 142: this function is now _almost_ just a wrapper of `findContours` which isn't terrible, but, since it is only called once, you might save some code and tests just by getting rid of it and including your logic in the calling function.

isaak-willett:
binary_mask
edge_mask
dilated_mask
Mask images produced

PR #26, opened 2020-04-14
Feature/add setup.py

djkapner:
This repo needed a `setup.py`.
I typically run my tests with `python setup.py test` where `setup.py` has a test command. For a while now, I've been getting this warning:
```
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
```
So I decided to look at tox and figure it out. This is my tox 101 implementation. I'm not sure this is the best way to go about it, or if we even want it. Any suggestions for a better way are welcome.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,7 @@
+[tox]
+envlist = py38
+[testenv]
+deps = 
+    -rrequirements.txt
+    pytest
+commands = pytest
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,0 @@
-[tox]
-envlist = py38
-[testenv]
-deps = 
-    -rrequirements.txt
-    pytest
-commands = pytest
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -0,0 +1,7 @@
+[tox]
+envlist = py38
+[testenv]
+deps = 
+    -rrequirements.txt
```
kschelonka, line 5: looks like a typo here

djkapner: I don't think it is:
https://tox.readthedocs.io/en/latest/example/basic.html#depending-on-requirements-txt-or-defining-constraints

kschelonka:
I'll preface this by saying I haven't used tox. That said, I don't think that this is the proper way to integrate tox with circleci. It will push everything onto one build, which starts with python3.8 (but then goes through whatever we have configured for the tox workflow -- so reinstalling python and all the dependencies). We won't benefit from the cache, and the different environments can't be parallelized.

I don't see a strong need to test multiple environments locally, and it is very easy to configure circle to run additional python environments in parallel. I'd need to be convinced for why we should add tox.

PR #27, opened 2020-04-15
S3 upload utilities and pipeline module

djkapner:
Creating a PR for early feedback on S3 upload utility functions and moto3 test usage.
The actual upload pipeline/module is still in dev.


tests/test_transfer_utils.py
```diff
@@ -0,0 +1,130 @@
...
+def db_file(tmpdir_factory):
+    db = tmpdir_factory.mktemp("upload").join("mydatabase.db")
+    conn = sqlite3.connect(db)
+    conn.execute("CREATE TABLE manifest_table (manifest text)")
+    conn.close
+    yield db
+
+
+@pytest.mark.parametrize("key", ["abc/temp.csv", None])
+def test_upload_file(local_file, bucket, key):
```
kschelonka, line 51: You can either use @mock_s3 or start the mock s3 service in the test to mock out all function calls to s3.

djkapner: It does work:
https://circleci.com/gh/AllenInstitute/segmentation-labeling-app/119

segmentation_labeling_app/transfers/utils.py
```diff
@@ -0,0 +1,188 @@
...
+    s3_uri : str
+        URI to S3 object just uploaded
+
+
+    """
+    if key is None:
+        key = pathlib.PurePath(file_name).name
+
+    client = boto3.client('s3')
+    client.upload_file(file_name, bucket, key)
```
djkapner, line 34: As per slack discussion, it is probably best practice to include a round-trip download and contents check here.

segmentation_labeling_app/transfers/upload.py
```diff
@@ -0,0 +1,94 @@
...
+        missing="",
+        description="key prefix for manifest contents")
+    manifest_prefix = argschema.fields.Str(
+        required=False,
+        allow_none=True,
+        default=None,
+        missing=None,
+        description=("key prefix for manifest. Will default "
+                     "to contents_prefix"))
+    manifest_write_mode = argschema.fields.Str(
```
djkapner, line 36: NOTE: as per slack discussion, we probably do not need to support appending a manifest on s3. This option and the supporting code can probably be dropped.

segmentation_labeling_app/transfers/utils.py
djkapner: included roundtrip check

segmentation_labeling_app/transfers/upload.py
djkapner: removed option to append to existing manifest

tests/test_transfer_utils.py
kschelonka: Cool, didn't know that. Thanks!

PR #31, opened 2020-04-16
Visualization evaluation

kschelonka:
Create simple jupyter notebook with widget for visualizing ROIs from the database. Meant to view and select based on size and shape, i.e. for good examples for training. The notebook has to be run somewhere with connection to the database to work.

Added methods to generate a "postage stamp" of the ROI. Can specify a canvas size. If it's larger than the ROI, will pad the canvas size with the ROI centered. Note that this is a "naive" center -- the center of the minimal rectangle that contains all the coordinates.

Screenshot of the notebook:
Screen Shot 2020-04-16 at 2 30 44 PM

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,6 @@
 pytest
 boto3
 moto
+numpy
+scipy
+ipywidgets
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,6 @@ moto
 numpy
 scipy
 ipywidgets
+matplotlib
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -1,3 +1,8 @@
 pytest
 boto3
 moto
+numpy
+scipy
+ipywidgets
+matplotlib
+psycopg2-binary
```
djkapner, line 8: need pandas

PR #33, opened 2020-04-22
Feature/roi refactor

djkapner:
refactor motivation:
* was writing the transform pipeline and needed additional functionality. Kat had added some useful utility functions in `rois/array_utils`
* added more sophisticated quantile, in addition to absolute thresholding. This will be better because ROIs don't have a consistent weight range.
* wanted a clearer interface to the design-specified functionality for the ROI class.
* fixed some postgres query issues.

With these changes, sizing masks and outlines to be the same shape and center will be enabled easily, for example:
```
mask = roi.generate_ROI_mask(
    shape=self.args['cropped_shape'])
outline = roi.generate_ROI_outline(
    shape=self.args['cropped_shape'],
    cumulative_threshold=self.args['outline_threshold'])
```

Some of the mask and outline functionality demonstrated:
```
r = rois.ROI.roi_from_query(1972)
mask = r.generate_ROI_mask()
plt.imshow(mask)        
```
image

```
mask = r.generate_ROI_mask(full=True)
plt.imshow(mask)        
```
image

```
mask = r.generate_ROI_mask(shape=(128, 128))
plt.imshow(mask)        
```
image

```
outline = r.generate_ROI_outline()
plt.imshow(outline, cmap='gray')
```
image

```
outline = r.generate_ROI_outline(full=True)
plt.imshow(outline, cmap='gray')
```
image

```
outline = r.generate_ROI_outline(shape=(128, 128))
plt.imshow(outline, cmap='gray')
```
image

```
outline = r.generate_ROI_outline(
    shape=(128, 128),
    quantile=0.01,
    inner_outline=True,
    dilation_kernel_size=4)
plt.imshow(outline, cmap='gray')
```
image

### DIFF
--- a/segmentation_labeling_app/rois/rois.py
+++ b/segmentation_labeling_app/rois/rois.py
@@ -1,14 +1,15 @@
 from typing import List, Tuple, Union
 import json
-import logging
 from pathlib import Path
-
+from scipy.optimize import bisect
 from scipy.sparse import coo_matrix
 import numpy as np
 import cv2
 import sqlite3
 
 import segmentation_labeling_app.utils.query_utils as query_utils
+from segmentation_labeling_app.transforms.array_utils import (
+        center_pad_2d, crop_2d_array)
 
 
 sql_create_rois_table = """ CREATE TABLE IF NOT EXISTS rois_prelabeling (
@@ -24,8 +25,110 @@ sql_create_rois_table = """ CREATE TABLE IF NOT EXISTS rois_prelabeling (
 sql_insert_roi = """ INSERT INTO rois_prelabeling (transform_hash,
                      ophys_segmentation_commit_hash,
                      creation_date, upload_date, manifest, experiment_id,
-                     roi_id) 
+                     roi_id)
                      VALUES (?, ?, ?, ?, ?, ?, ?) """
+
+
+def cumulative_fraction_threshold(data, target_fraction, nbins=200):
+    """return a threshold value above which the summed weights are
+    target_fraction of the total summed weights
+
+    Parameters
+    ----------
+    data : numpy.ndarray
+        1d array of data, all values > 0
+    target_fraction : float
+        in range [0, 1.0]
+
+    Returns
+    -------
+    fval : float
+        data[data > fval].sum() / data.sum() ~ target_fraction
+
+    """
+    def fopt(x):
+        frac = data[data > x].sum() / data.sum()
+        diff = target_fraction - frac
+        return diff
+
+    fval = bisect(fopt, data.min(), data.max())
+
+    return fval
+
+
+def binary_mask_from_threshold(
+            arr: Union[np.ndarray, coo_matrix],
+            absolute_threshold: float = None,
+            cumulative_threshold: float = 0.9,
+            nbins: int = 50) -> np.array:
+    """Binarize an array
+
+    Parameters
+    ----------
+    arr: numpy.ndarray or scipy.sparse.coo_matrix
+        2D array of weighted floating-point values
+    absolute_threshold: float
+        weighted entries above this value=1, below=0
+        over-ridden if cumulative_threshold is set
+    cumulative_threshold: float
+        if specified, set absolute threshold determined by fraction of
+        total weight above this value
+    nbins: int
+        if cumulative specified, determines
+        for interpolating
+
+    Returns
+    -------
+    binary: numpy.ndarray
+        binarized mask
+
+    """
+    wmask = arr
+    if isinstance(arr, coo_matrix):
+        wmask = arr.toarray()
+    vals = wmask[np.nonzero(wmask)]
+
+    if cumulative_threshold is not None:
+        absolute_threshold = cumulative_fraction_threshold(
+                vals,
+                cumulative_threshold,
+                nbins=nbins)
+
+    binary = np.uint8(wmask > absolute_threshold)
+
+    return binary
+
+
+def sized_mask(
+        arr: Union[np.ndarray, coo_matrix], shape: Tuple[int, int] = None,
+        full: bool = False):
+    """return a 2D dense array representation of the mask, optionally
+    cropped and padded
+
+    Parameters
+    ----------
+    arr: numpy.ndarray or scipy.sparse.coo_matrix:
+        a representation of the mask
+    shape: tuple(int, int)
+        [h, w] for padded shape. If None, cropped to existing values
+    full: bool
+        if True, the full-frame array is returned
+
+    Returns
+    -------
+    mask: numpy.ndarray
+        2D dense matrix representation of the mask
+
+    """
+    if isinstance(arr, coo_matrix):
+        mask = arr.toarray()
+    else:
+        mask = arr
+    if not full:
+        mask = crop_2d_array(mask)
+        if shape is not None:
+            mask = center_pad_2d(mask, shape)
+    return mask
 
 
 class ROI:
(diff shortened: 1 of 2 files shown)


segmentation_labeling_app/rois/rois.py
```diff
@@ -24,10 +25,112 @@
 sql_insert_roi = """ INSERT INTO rois_prelabeling (transform_hash,
                      ophys_segmentation_commit_hash,
                      creation_date, upload_date, manifest, experiment_id,
-                     roi_id) 
+                     roi_id)
                      VALUES (?, ?, ?, ?, ?, ?, ?) """
 
 
+def cumulative_fraction_threshold(data, target_fraction, nbins=200):
```
isaak-willett, line 32: Is nbins not used in this function? It's also not in the docstring.

segmentation_labeling_app/rois/rois.py
```diff
@@ -24,10 +25,112 @@
...
+    """return a threshold value above which the summed weights are
+    target_fraction of the total summed weights
+
```
isaak-willett, line 35: Just for clarification this function returns a value such that the percent of values above the found threshold is equal to the specified target_fraction? Am I understanding that right?

segmentation_labeling_app/rois/rois.py
```diff
@@ -24,10 +25,112 @@
...
+    fval : float
+        data[data > fval].sum() / data.sum() ~ target_fraction
+
+    """
+    def fopt(x):
+        frac = data[data > x].sum() / data.sum()
+        diff = target_fraction - frac
+        return diff
+
+    fval = bisect(fopt, data.min(), data.max())
```
isaak-willett, line 54: The [docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.bisect.html) say this can return a root results object when full_output is true, would it beneficial to have that for testing for convergence and not assuming convergence.

segmentation_labeling_app/rois/rois.py
```diff
@@ -92,102 +195,86 @@ def roi_from_query(cls, roi_id: int) -> "ROI":
...
-        if len(edge_coordinates) > 0:
-            for edge_coordinate in edge_coordinates:
-                mask_matrix[edge_coordinate[1]][edge_coordinate[0]] = 1
-        return mask_matrix
+        mask = np.zeros_like(binary)
+        mask[xy[:, 1], xy[:, 0]] = 1.0
+
+        kernel = np.ones((dilation_kernel_size, dilation_kernel_size))
+
+        mask = cv2.dilate(mask, kernel)
```
isaak-willett, line 270: Do we want to specify iterations on the dilation? If dilation iterations are specified the bitwise and will not work for subtracting away non outline mask pixels.

tests/test_rois.py
```diff
@@ -1,209 +1,239 @@
...
-    generated_mask = test_roi.generate_binary_mask_from_threshold(threshold=threshold)
-    assert np.array_equal(generated_mask, expected_mask)
+@pytest.mark.parametrize(
+        ("data", "target_fraction", "expected"),
+        [
+            (np.linspace(0, 1.0, 1000), 0.95, 0.95),
+            (np.linspace(0, 1.0, 1000), 0.05, 0.05),
+            (np.array([1.0, 1.0, 5.0, 5.0]), 0.65, 10/12)
+            ])
+def test_cumulative_fraction(data, target_fraction, expected):
```
isaak-willett, line 16: I would consider adding the edge case of non convergent fraction, ie: what if all the data values are 0.7 and the requested fraction is 0.5.

tests/test_rois.py
```diff
@@ -1,209 +1,239 @@
...
+                ])
+def test_binary_mask_from_threshold(
+        weighted, expected, athresh, cthresh, use_coo):
+    if use_coo:
+        weighted = coo_matrix(weighted)
+    binary = roi_module.binary_mask_from_threshold(
+            weighted,
+            absolute_threshold=athresh,
+            cumulative_threshold=cthresh)
+    print(binary)
```
isaak-willett, line 62: unneeded print statement

tests/test_rois.py
```diff
@@ -1,209 +1,239 @@
...
+                    [0, 0, 1, 0, 0],
+                    [0, 1, 0, 1, 0],
+                    [1, 1, 1, 0, 0],
+                    [0, 0, 0, 0, 0]
+                    ]),
+                0.7,
+                None,
+                True)])
+def test_roi_generate_outline(weighted, full, expected, athresh, cthresh):
+    """this test is not exhaustive
```
isaak-willett, line 225: I would consider adding a test case where cv contours finds multiple contours for one ROI, this should be handled in your function by np.concatinate(contours) but I think it's an important test to have.

segmentation_labeling_app/rois/rois.py
djkapner: right. left over from a previous attempt. removing.

segmentation_labeling_app/rois/rois.py
djkapner: I think this part of the docstring is unambiguous:
```
    Returns
    -------
    fval : float
        data[data > fval].sum() / data.sum() ~ target_fraction
```

segmentation_labeling_app/rois/rois.py
djkapner: It is definitely possible we're going to run into some weird cases. But, we just want to get something running for UI/UX testing. Let's come back and handle weird cases later.

segmentation_labeling_app/rois/rois.py
djkapner: I think the default of 1 is fine. This is mainly for visualization.
I think the AND is working

tests/test_rois.py
djkapner: right. removed

tests/test_rois.py
djkapner: I'm not sure what you mean here. It is only ever passed 1 ROI. That's what the concatenate is for.

segmentation_labeling_app/rois/rois.py
isaak-willett: I agree, it was for my own clarification

segmentation_labeling_app/rois/rois.py
isaak-willett: For clarification, the AND won't work with iterations > 1, as we expand away from the original mask it will fail where we have expanded the borders. But that's not relevant if we're capping the iterations at 1.

tests/test_rois.py
isaak-willett: I had cases where one ROI was segmented into multiple contours, ie: one ROI had four or five indexes in the upper level list as cv2 said they weren't part of the same shape. This was found during testing on the ROIs table. The concatenate should catch this but for complete test coverage I think it's a good idea to include so we have record. If we're adding tests later however it can probably wait but it's a good test case.

segmentation_labeling_app/rois/rois.py
djkapner: still don't know what you mean. This is what I would expect:
```
In [220]: binary                                                                   
Out[220]: 
array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]], dtype=uint8)

In [221]: contours, _ = cv2.findContours(binary, cv2.RETR_LIST, cv2.CHAIN_APPROX_NO
     ...: NE)                                                                      

In [222]: xy = np.concatenate(contours).squeeze()                                  

In [223]: mask = np.zeros_like(binary)                                             

In [224]: mask[xy[:, 1], xy[:, 0]] = 1.0                                           

In [225]: mask                                                                     
Out[225]: 
array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 0, 0, 0, 0, 0, 0, 1, 0],
       [0, 1, 0, 0, 0, 0, 0, 0, 1, 0],
       [0, 1, 0, 0, 0, 0, 0, 0, 1, 0],
       [0, 1, 0, 0, 0, 0, 0, 0, 1, 0],
       [0, 1, 0, 0, 0, 0, 0, 0, 1, 0],
       [0, 1, 0, 0, 0, 0, 0, 0, 1, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]], dtype=uint8)

In [226]: mask = cv2.dilate(mask, kernel)                                          

In [227]: mask                                                                     
Out[227]: 
array([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 1, 0, 0, 0, 0, 1, 1, 1],
       [1, 1, 1, 0, 0, 0, 0, 1, 1, 1],
       [1, 1, 1, 0, 0, 0, 0, 1, 1, 1],
       [1, 1, 1, 0, 0, 0, 0, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]], dtype=uint8)

In [228]: mask = mask & binary                                                     

In [229]: mask                                                                     
Out[229]: 
array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 1, 0, 0, 0, 0, 1, 1, 0],
       [0, 1, 1, 0, 0, 0, 0, 1, 1, 0],
       [0, 1, 1, 0, 0, 0, 0, 1, 1, 0],
       [0, 1, 1, 0, 0, 0, 0, 1, 1, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]], dtype=uint8)
```

tests/test_rois.py
djkapner: I expect the multiple elements of the contour list. That's why I added a `np.concatenate()`
There's nothing wrong here with us getting multiple contours back.

segmentation_labeling_app/rois/rois.py
kschelonka: What's the reason for this method vs. just computing quantiles (e.g. https://docs.scipy.org/doc/numpy/reference/generated/numpy.quantile.html)

segmentation_labeling_app/rois/rois.py
kschelonka: Isaak I think the source of confusion here could be that Dan is dilating the mask, not the outline.

isaak-willett: I expect the same thing to happen as what you posted with the AND. Unless I'm mistaken the borders are supposed to expand outward with multiple iterations not inwards towards the center, so the AND subtracts away those border pixels. What I would expect from the dilation_iterations==2 would be the mask below, which the AND function does not provide us.
```
      [[1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 0, 0, 0, 0, 0, 0, 1, 1],
       [1, 1, 0, 0, 0, 0, 0, 0, 1, 1],
       [1, 1, 0, 0, 0, 0, 0, 0, 1, 1],
       [1, 1, 0, 0, 0, 0, 0, 0, 1, 1],
       [1, 1, 0, 0, 0, 0, 0, 0, 1, 1],
       [1, 1, 0, 0, 0, 0, 0, 0, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]]
```
I think this is the expansion direction because otherwise the border would begin to overlap the ROI and it would be harder for the labeler to see. But correct me if I'm wrong.

segmentation_labeling_app/rois/rois.py
djkapner: No, that's pretty much what I was trying to do, I just didn't know the `quantile` function

tests/test_rois.py
isaak-willett: I agree there's nothing wrong, and i agree with the concatenate. It was an bug I earlier not using the concatenate and thought it would be a good test case to keep in there. But if we're aware of it letting it pass is fine with me.

segmentation_labeling_app/rois/rois.py
djkapner: Ah, I see the misunderstanding. `dilate` doesn't know anything about inside or outside. It just dilates:
https://docs.opencv.org/4.2.0/d4/d86/group__imgproc__filter.html#ga4ff0f3318642c4f469d0e11f242f3b6c

tests/test_rois.py
kschelonka: I agree with Isaak here. An example is if the soma is missing from the ROI, so there's a large "hole" in the middle. You'd expect 2 outlines -- one for the hole and one for the outside.

segmentation_labeling_app/rois/rois.py
isaak-willett: Right, and we would be expanding our borders inward with each iteration using the AND not outwards. This was the motivation for the iterative egde detection and dilating in my previous ROIs class, I didn't know a way to do that process more efficiently.

djkapner: I'm still not sure what the complaint is here. Again, this is just for visualization, but, what Kat wanted was thicker borders that don't extend outwards past the original ROI boundary. She wanted the "inner outline"

tests/test_rois.py
```diff
@@ -1,209 +1,238 @@
...
-        3) 0 x 0 object in a 5 x 5 frame, edge mask gets nothing
-        4) 1 x 1 object in a 5 x 5 frame, edge mask return 1 x 1 object
-    """
-    test_roi = roi_module.ROI(coo_rows, coo_cols, coo_data,
-                              video_shape, segmentation_id, roi_id)
-    edge_mask = test_roi.get_edge_mask(threshold=threshold)
-    assert np.array_equal(expected_mask, edge_mask)
+@pytest.mark.parametrize("use_coo", [True, False])
+@pytest.mark.parametrize("mask, full, shape, expected", [
+    (
```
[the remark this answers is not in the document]
djkapner: will do

tests/test_rois.py
djkapner: I see, you want the test case where you explicitly see multiple contours. We're quite often getting multiple contours even in a not-so-obvious example as that. But, sure.

segmentation_labeling_app/rois/rois.py
isaak-willett: @kschelonka I was under the impression the borders were not supposed to overlap the ROI so they should expand outward from the 'inner outline'. In my mind if they expand inward the outline will overlap the ROI which will lead to a bad labeling experience. Could you clarify for me which way the ROI edges are to expand? Thanks!

djkapner: oh, if I misunderstood that, we can just change to `mask = mask & (not binary)`

isaak-willett: I think we still want the original border and if there's a hole that will add the entire hole to the mask, if I'm not mistaken. But I could be wrong, I'll go and test that and see if it comes across okay.

segmentation_labeling_app/rois/rois.py
djkapner: Thanks, that really cleans things up. Redoing the plots in the PR comments after this change.

djkapner: PR comment updated.

segmentation_labeling_app/rois/rois.py
kschelonka: Dan is correct about what I meant by "inner outline". My thought originally was that we want the outline to be bounded to the actual edge of the cell, and not extend past it. If we generate a few examples of each way I'm open to considering an outline that would extend beyond the cell. Since the outlines are not going to be very thick it probably won't matter a great deal.

kschelonka: The user will have the option to view a transparent mask overlay or an outline.

tests/test_rois.py
djkapner: working on adding this

segmentation_labeling_app/rois/rois.py
isaak-willett: That's my bad for the confusion, still confusion over which way we are expanding lol. I think we should expand outward as we don't want to intersect the actual ROI, I think it'll make it hard to view. If you both are on the same page about borders expanding inward I'm okay with that.

tests/test_rois.py
djkapner: done

tests/test_rois.py
```diff
@@ -1,209 +1,257 @@
...
+                    [1, 0, 1, 1, 0, 1],
+                    [1, 0, 1, 1, 0, 1],
+                    [1, 0, 0, 0, 0, 1],
+                    [1, 1, 1, 1, 1, 1]
+                    ]),
+                0.7,
+                None,
+                True),
+            ])
+def test_roi_generate_outline(weighted, full, expected, athresh, quantile):
```
isaak-willett, line 239: I think as with above comment, adding descriptions for these test cases would be helpful.

djkapner: added

PR #34, opened 2020-04-22
Feature/update array utils

djkapner:
This PR should merge after #33

NOTE: I don't feel like the code in `rois/array_utils.py` is optimal, and I don't think the test coverage for this new function is complete. But, the basic functionality I want is there, and I want to merge and move on to finishing the pipeline and come back to this when corner cases come up.

Purpose of the function `content_extents`:
* given an ROI object, give back some extents and padding to index and pad a video frame

Example:
image
from:
```
import segmentation_labeling_app.rois.rois as rois
import segmentation_labeling_app.transforms.array_utils as au
import matplotlib.pyplot as plt
import numpy as np

# get an ROI and its mask
roi = rois.ROI.roi_from_query(1972)
mask = roi.generate_ROI_mask()

# make a moview frame with context around the ROI
x = np.arange(512)
z = np.meshgrid(np.sin(x/4), np.cos(x/4))
frame = np.abs(z[0] + z[1]) * 0.4
frame += roi._sparse_coo.toarray()

# use the new function to extract the part of the movie
# frame based on the ROI
e, pad = au.content_extents(roi._sparse_coo, shape=(128, 128), target_shape=frame.shape)
frame_extract = np.pad(frame[e[0]: e[1], e[2]:e[3]], pad)

f, a = plt.subplots(1, 3, clear=True, num=1)
a[0].imshow(mask)
a[0].set_title('mask')
a[1].imshow(frame)
a[1].set_title('movie frame')
a[2].imshow(frame_extract)
a[2].set_title('extracted region of frame')
```

changing the roi placement:
```
roi = rois.ROI.roi_from_query(1972)
roi._sparse_coo.row -= 60
```
image

and again:
```
roi = rois.ROI.roi_from_query(1972)
roi._sparse_coo.row += 400
roi._sparse_coo.col += 370
```
image

PR #37, opened 2020-04-23
Feature/transform pipeline

djkapner:
NOTE: absorbs now defunct #34. Those PR comments are copied below.

Missing items:
- [x] trace transform (no transform provided yet)
- [x] write out per-ROI manifest to ~sqlite~ postgres
- [x] change from sqlite to 3rd postgres table (nice-to-have)
- [x] output schema for the pipeline class.
- [x] tests of pipeline, mock out all transformation calls so that this is testing only the file writing and manifest generation.

NOTE: For faster testing and not having to read in an entire movie. `segmentation_run_id=953`is a special test case with ophys commit hash `truncated_test_movie`. The ROIs in rois table for the original movie were copied, and key to this entry.
I am testing with this example call.
```
python -m segmentation_labeling_app.transforms.transform_pipeline \
    --segmentation_run_id 950 \
    --artifact_basedir /allen/aibs/informatics/labeling_artifacts/
```
which created this dir `/allen/aibs/informatics/labeling_artifacts/segmentation_run_id_950` and filled it with tons of pngs and mp4s.

#34 comments:
NOTE: I don't feel like the code in `rois/array_utils.py` is optimal, and I don't think the test coverage for this new function is complete. But, the basic functionality I want is there, and I want to merge and move on to finishing the pipeline and come back to this when corner cases come up.

Purpose of the function `content_extents`:
* given an ROI object, give back some extents and padding to index and pad a video frame

Example:
image
from:
```
import segmentation_labeling_app.rois.rois as rois
import segmentation_labeling_app.transforms.array_utils as au
import matplotlib.pyplot as plt
import numpy as np

# get an ROI and its mask
roi = rois.ROI.roi_from_query(1972)
mask = roi.generate_ROI_mask()

# make a moview frame with context around the ROI
x = np.arange(512)
z = np.meshgrid(np.sin(x/4), np.cos(x/4))
frame = np.abs(z[0] + z[1]) * 0.4
frame += roi._sparse_coo.toarray()

# use the new function to extract the part of the movie
# frame based on the ROI
e, pad = au.content_extents(roi._sparse_coo, shape=(128, 128), target_shape=frame.shape)
frame_extract = np.pad(frame[e[0]: e[1], e[2]:e[3]], pad)

f, a = plt.subplots(1, 3, clear=True, num=1)
a[0].imshow(mask)
a[0].set_title('mask')
a[1].imshow(frame)
a[1].set_title('movie frame')
a[2].imshow(frame_extract)
a[2].set_title('extracted region of frame')
```

changing the roi placement:
```
roi = rois.ROI.roi_from_query(1972)
roi._sparse_coo.row -= 60
```
image

and again:
```
roi = rois.ROI.roi_from_query(1972)
roi._sparse_coo.row += 400
roi._sparse_coo.col += 370
```
image

## Running the pipeline
(the following is checing out this feature branch. In production, we will eliminate the checkout to get the default branch.)
```
#!/bin/bash
#PBS -q braintv
#PBS -l nodes=1:ppn=5
#PBS -l mem=50g
#PBS -l walltime=01:00:00
#PBS -N prelabel_transforms
#PBS -j oe
#PBS -o /allen/aibs/informatics/danielk/segmentation_results

export TMPDIR=/scratch/capacity/${PBS_JOBID}

module load anaconda
conda create --prefix ${TMPDIR}/myconda python=3.7
source activate ${TMPDIR}/myconda
cd ${TMPDIR}
git clone https://github.com/AllenInstitute/segmentation-labeling-app
cd segmentation-labeling-app
git checkout feature/transform_pipeline
export TRANSFORM_HASH=$(git rev-parse HEAD)
pip install .

source /allen/aibs/informatics/danielk/segmentation_results/sourceme.sh

python -m segmentation_labeling_app.transforms.transform_pipeline \
    --segmentation_run_id 954 \
    --artifact_basedir /allen/aibs/informatics/labeling_artifacts/
```
produced entries in the postgres table `roi_manifests`:
image

and the following assertions on those rows pass:
```
import segmentation_labeling_app.utils.query_utils as qu                  
import os 
                                                                
lv = qu.get_labeling_db_credentials()                                     
db_conn = qu.DbConnection(**lv)                                           
mans = db_conn.query( 
    "SELECT rm.* FROM roi_manifests AS rm " 
    "JOIN rois on rois.id=rm.roi_id " 
    "WHERE rois.segmentation_run_id=954")                                 

for m in mans: 
    mjson = m['manifest'] 
    for k in mjson.keys(): 
        if k not in ['experiment-id', 'roi-id']:
            assert os.path.isfile(mjson[k])                                          
```


kschelonka:
What do we need for the trace transform? Is it not possible to use the existing traces?

djkapner:
> What do we need for the trace transform? Is it not possible to use the existing traces? I understand we'll need to load them and convert them to json for upload to s3.

I didn't write them out anywhere for populating the postgres rois table. I skipped the step where we do Allen trace extraction (for speed) and I did not save Suite2P outputs (for space). This was because I was [expecting a trace transform](https://github.com/AllenInstitute/informatics_design_docs/tree/master/ophys/segmentation#generate-trace-transform).

Suite2P traces might be best, as they will be demixed. Our simple Allen thing did not demix. It isn't that bad to rerun the ROIs segmentations. Maybe a day. So, if we want to capture the Suite2P traces or something inside of `ophys_segmentation` we could do that.

kschelonka:
Thanks for your comments. Is there anything else in the transformations design doc that has not yet been implemented?

segmentation_labeling_app/transforms/array_utils.py
```diff
@@ -44,6 +44,89 @@ def content_boundary_2d(arr: Union[np.ndarray, coo_matrix]) -> np.ndarray:
...
+    -------
+    indexing_bounds: tuple
+        4-tuple of row/column boundaries
+    pad_width: tuple(tuple, tuple)
+        to be passed into numpy.pad as pad_width
+    target_shape: (Tuple[int, int]) Extent of array to be indexed. If None
+        padding will still handle top and left, but bottom and right padding
+        will always be zero
+
+    Returns
```
kschelonka, line 76: Duplicated Returns section of your docstring here

segmentation_labeling_app/transforms/array_utils.py
```diff
@@ -44,6 +44,89 @@ def content_boundary_2d(arr: Union[np.ndarray, coo_matrix]) -> np.ndarray:
     return top_bound, bot_bound, left_bound, right_bound
 
 
+def content_extents(
```
kschelonka, line 47: What differentiates this from `center_pad_2d`?

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -91,105 +91,107 @@ def _downsample_array(full_array: h5py.Dataset, input_fps: int,
...
-        projection_type: maximum or average, what type of projection
-
-    Returns: A numpy array generated with the specified strategy (time, row, col)
-
-    """
-    if projection_type == 'average':
-        return np.mean(video)
-    elif projection_type == 'maximum':
-        return np.max(video, axis=0)
+# def get_transformed_center(coordinate_pair: Tuple[int, int],
```
kschelonka, line 94: If this is no longer required, I think we should delete rather than comment.

segmentation_labeling_app/transforms/array_utils.py
djkapner: `center_pad_2d` pads an array, for example a cropped ROI, with zeros. But, I want to know from this the indices that would crop it out of the original video frame. You can see its use in `TransformPipeline`

segmentation_labeling_app/transforms/transformations.py
djkapner: agreed, just waiting to see where chips fall

segmentation_labeling_app/transforms/array_utils.py
djkapner: fixed.

segmentation_labeling_app/transforms/array_utils.py
djkapner: I copied in the images (see PR description comment above) from PR #34 (now absorbed into this one). You can't index a movie with `center_pad_2d` the way I wanted to in the demo.

segmentation_labeling_app/transforms/transform_pipeline.py
```diff
@@ -0,0 +1,159 @@
...
+                raise TransformPipelineException(
+                    f"{query_string} did not return exactly 1 result")
+            data['segmentation_run_id'] = entries[0]['id']
+        return data
+
+
+class TransformPipeline(argschema.ArgSchemaParser):
+    default_schema = TransformPipelineSchema
+
+    def run(self):
```
isaak-willett, line 84: Could we separate the logic of the run function into a non class method like in suite2p_job. I think it makes the unit testing more concise and easier to read when this is the case.

djkapner: I suppose we could, but this is a pipeline that runs a bunch of unit-tested functions. Running test code on this class is pretty much an integration test. Given that, I don't see the advantage to your suggestion. Does what I am saying make sense, or could you explain more?

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -226,4 +228,3 @@ def transform_to_mp4(video: np.ndarray, output_path: str,
     for frame in norm_video:
         writer.send(frame)
     writer.close()
-
```
isaak-willett, line 229: If this is the last line I think pep8 says you need a line delimiter at the end of a file

segmentation_labeling_app/utils/query_utils.py
```diff
@@ -54,4 +56,4 @@ def query(query, user, host, database, password, port):
     finally:
         cursor.close()
         conn.close()
-    return results
\ No newline at end of file
+    return results
```
isaak-willett, line 59: Again I think there should be a new line delimiter at the end of the file

tests/test_transformations.py
```diff
@@ -86,33 +86,33 @@ def test_video_downsampling(input_fps, output_fps, random_seed, strategy,
...
-                                                                    video_shape)
-        assert transformed_center == expected_pair
+# @pytest.mark.parametrize(("coordinate_pair, window_size, video_shape,"
+#                           "expected_pair"), [
+#                         ((0, 9), (10, 10), (512, 512), (4, 9)),
+#                         ((9, 0), (10, 10), (512, 512), (9, 4)),
+#                         ((511, 4), (10, 10), (512, 512), (506, 4)),
+#                         ((4, 511), (10, 10), (512, 512), (4, 506)),
+#                         ((5, 5), (3, 3), (5, 5), (3, 3)),
+#                         ((-1, 2), (3, 3), (5, 5), (1, 2))])
```
isaak-willett, line 96: I would consider adding a test case such that coordinate_pair (+,-) for complete coverage.

segmentation_labeling_app/transforms/transformations.py
djkapner: I agree, we should all be striving for pep8. FYI, here's the linting output on the master branch for this file. Please start checking this in some way in your code submissions.
```
flake8 segmentation_labeling_app/transforms/transformations.py 
segmentation_labeling_app/transforms/transformations.py:25:80: E501 line too long (80 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:33:80: E501 line too long (83 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:99:80: E501 line too long (80 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:101:80: E501 line too long (83 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:102:80: E501 line too long (80 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:110:80: E501 line too long (82 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:124:80: E501 line too long (81 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:137:80: E501 line too long (82 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:141:80: E501 line too long (82 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:155:80: E501 line too long (80 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:171:80: E501 line too long (88 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:186:80: E501 line too long (81 > 79 characters)
segmentation_labeling_app/transforms/transformations.py:229:1: W391 blank line at end of file
```

tests/transforms/test_array_utils.py
```diff
@@ -132,3 +132,105 @@ def test_crop_array_raises_error(arr):
...
+                    [0, 0, 0, 0, 0, 0, 0],
+                    [0, 0, 0, 0, 0, 0, 0],
+                    [0, 0, 1, 1, 1, 0, 0],
+                    [0, 0, 1, 2, 2, 0, 0],
+                    [0, 0, 1, 1, 3, 0, 0],
+                    [0, 0, 0, 0, 0, 0, 0],
+                    [0, 0, 0, 0, 0, 0, 0]]),
+                (5, 5)),
+            ])
+def test_compare_crop_pad_and_extents(arr, shape):
```
isaak-willett, line 226: I think this would be cleaner providing the expected array as a parameter rather than creating the indexed array in the test function

segmentation_labeling_app/transforms/transformations.py
djkapner: I think if you try it, you will see that there is a line delimiter at the end of the file, there is not a blank line though. These are different.

segmentation_labeling_app/utils/query_utils.py
djkapner: I think if you try it, you will see that there is a line delimiter at the end of the file, there is not a blank line though. These are different.

isaak-willett: Right that's my bad I didn't see the delimiter.

segmentation_labeling_app/transforms/transformations.py
isaak-willett: I agree with pep8 other than line length, I think it makes the code harder to read if there's a hard line at 80 chars but that's a discussion for slack not code comments hah. I see the delimiter that's my mistake.

segmentation_labeling_app/transforms/transform_pipeline.py
isaak-willett: I think writing that integration test is cleaner when they are separated is what I'm getting at. I think it's easier to understand the test code. But if we don't intend for this to be fully integration tested then I'm not opposed to leaving it as is.

tests/transforms/test_array_utils.py
```diff
@@ -132,3 +132,105 @@ def test_crop_array_raises_error(arr):
...
+    """check that cropping and padding an image
+    is the same as getting the extents and
+    indexing the array (i.e. as would be applied to video frame)
+    """
+    cropped_padded = au.center_pad_2d(au.crop_2d_array(arr), shape)
+    extents, pad_width = au.content_extents(arr, shape)
+    indexed = np.pad(
+            arr[extents[0]:extents[1], extents[2]:extents[3]],
+            pad_width)
+    assert np.all(cropped_padded == indexed)
```
isaak-willett, line 236: I think np.testing.equal might be a better assertion function, either works

segmentation_labeling_app/transforms/transformations.py
isaak-willett: If it's working for your runs I think we should delete it in this PR.

segmentation_labeling_app/transforms/transform_pipeline.py
kschelonka: The meaningful integration testing here would be between the pipeline and the databases, and the pipeline outputs and the segmentation app. I don't think we'll get a lot of benefit from unit-testing `run` alone.

segmentation_labeling_app/transforms/array_utils.py
kschelonka: Hm, I see. So you want `content_boundary_2d` but to work for the 3d array?

kschelonka: I was just looking for places to consolidate since I see a lot of similar looking code. So if we can reuse more then it would clean it up a bit. I see now it's more of a nit.

djkapner: yes, that's right.
I agree, the two are very similar and could be combined. That was why I had said "code not optimal" in the original PR. But, these are touchy corner case things and didn't want to mess with this too much right now.

segmentation_labeling_app/transforms/transform_pipeline.py
djkapner: re: Isaak and cleaner test code. To test an ArgschemaParser class called `Something` like this you can simply do:
```
def test_something():
    args = {"a": 1, "b": 2}
    mysomething = Something(input_data=args, args=[])
    mysomething.run()
    assert(some_stuff)
```
I don't see how it is unclean. But, making the `run()` command just pass things to a function as function args eliminates the whole point of inheriting this class from ArgSchemaParser, and that was my design choice to make, IMO.

tests/transforms/test_array_utils.py
djkapner: the point of this test is that the two methods agree. That's what the comment is saying.

tests/transforms/test_array_utils.py
djkapner: That's not the name of a function. I assume you mean [`numpy.testing.assert_equal()`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_equal.html?highlight=assert_equal#numpy-testing-assert-equal)
If you have some experience or reason to share why you "think it might be a better assertion function" please share.

segmentation_labeling_app/transforms/transformations.py
djkapner: removed comment blocks.

tests/transforms/test_array_utils.py
isaak-willett: Yes thats the function name, my bad.
Assert_equal searches for the first instance of inequality and raises an exception then. It also prints the inequality of the arrays which to me produces much more meaningful stack traces than getting a true or false from np.all.

segmentation_labeling_app/transforms/transform_pipeline.py
isaak-willett: I think this is to taste but to me the use of argschema is to allow ease of use in the command line while maintaining usability in imports. I don't like creating objects if they aren't explicitly needed and I think that running this from another module shouldn't need to use argschema if it's not necessary. It's to taste to some extent but if that's your design choice that's fine, I just prefer as much functional style over OOP as possible.

segmentation_labeling_app/rois/rois.py
```diff
@@ -133,31 +137,28 @@ def roi_from_query(cls, roi_id: int) -> "ROI":
...
             ("SELECT * FROM segmentation_runs WHERE "
-             f"id={roi['segmentation_run_id']}"),
-            user=label_vars.user,
-            host=label_vars.host,
-            database=label_vars.database,
-            port=label_vars.port,
-            password=label_vars.password)[0]
+             f"id={roi['segmentation_run_id']}"))[0]
+
+        # NOTE: temporary legacy support for no traces
```
kschelonka, line 147: Can you please add `# TODO` to this? Using a standard makes it easy to search for later.

segmentation_labeling_app/transforms/transform_pipeline.py
```diff
@@ -0,0 +1,204 @@
...
+        required=False,
+        default=0,
+        description="random seed to use if downsampling strategy is 'random'")
+
+    @mm.pre_load
+    def set_segmentation_run_id(self, data, **kwargs):
+        """if segmentation_run_id not provided, query the database
+        using the other provided args
+        """
+        if "segmentation_run_id" not in data:
```
kschelonka, line 70: Is this alternative method easier, or is there another reason to include it? Seems like you'd have to query the segmentation_run_id table anyway to check the commit hash, unless you instead take the most recent run.

segmentation_labeling_app/transforms/transform_pipeline.py
```diff
@@ -0,0 +1,204 @@
...
+                     f"segmentation_run_id_{self.args['segmentation_run_id']}"
+        os.makedirs(output_dir, exist_ok=True)
+
+        # get all ROI ids from this segmentation run
+        query_string = ("SELECT id FROM rois WHERE segmentation_run_id="
+                        f"{self.args['segmentation_run_id']}")
+        entries = db_conn.query(query_string)
+        roi_ids = [i['id'] for i in entries]
+
+        # NOTE: here could be a good place to put a pre-filtering stepw
```
kschelonka, line 108: See above with `# TODO` for visibility

segmentation_labeling_app/transforms/transform_pipeline.py
djkapner: I was just trying to anticipate a different use case. We can take it out to keep it simple.

segmentation_labeling_app/rois/rois.py
djkapner: changed NOTEs to TODOs

segmentation_labeling_app/transforms/transform_pipeline.py
djkapner: changed NOTEs to TODOs

segmentation_labeling_app/transforms/transformations.py
```diff
@@ -1,195 +1,114 @@
...
-    Args:
-        coordinate_pair: the original coordinate location for centering (row, col)
-        box_size: the size of the box for subset (row, col)
-        video_shape: the x and y dimensions of the video (time, row, col)
+        rng = np.random.default_rng(random_seed)
 
-    Returns:
-        transformed_center: returns the best fitting center of the box for out
-        of bounds coordinates (row, col)
+        def sampler(arr, idx):
```
kschelonka, line 89: Nit:
Redefining `sampler` is a little confusing. Since these can all be done in one line, might be clearer to use a dictionary?

```
sampling_strategies = {
    "maximum": lambda arr, idx: arr[idx].max(axis=0),
    "random": arr[np.random.default_rng(random_seed).choice(idx)],
...
}
sampler = sampling_strategies[strategy]
for i, bin_indices in enumerate(bin_list):
    array_out[i] = sampler(array, bin_indices)
```

djkapner: yeah, I agree, this doesn't look nice. Initially I was naming lambdas for shorter code, but flake8 did not like this.I was expecting some good suggestions here. I like the looks of the dictionary, I'll do that.

kschelonka: Yeah, I see the purpose behind not naming lambdas but, I think it sometimes it could make things clearer. The dictionary is a good alternative (and also it's nice to select based on key).

tests/transforms/test_transform_pipeline.py
```diff
@@ -0,0 +1,135 @@
...
+         {"source_video_path": "/mock/path", "ophys_experiment_id": 123}
+      ]},
+
+     "/mock/path",
+
+     [{"experiment_id": 123, "roi_id": 0, "segmentation_run_id": 42},
+      {"experiment_id": 123, "roi_id": 777, "segmentation_run_id": 42}]),
+
+], indirect=["mock_db_conn_fixture"])
+def test_transform_pipeline(tmp_path, monkeypatch, mock_db_conn_fixture,
```
kschelonka, line 82: Wow, even testing the sql. Nice.

segmentation_labeling_app/transforms/transformations.py
djkapner: implemented

tests/transforms/test_transform_pipeline.py
djkapner: that @njmei right there

tests/transforms/test_transformations.py
```diff
@@ -0,0 +1,177 @@
...
+         [[4, 4], [4, 4]], [[0, 0], [0, 0]],
+         [[3, 2], [10, 7]], [[11, 1], [0, 5]],
+         [[8, 2], [12, 1]], [[6, 3], [2, 8]],
+         [[11, 3], [1, 17]], [[8, 3], [9, 0]]]
+    )
+    yield video
+
+
+@pytest.mark.parametrize(
+        ("array, input_fps, output_fps, random_seed, strategy, expected"),
```
kschelonka, line 25: Nice extension for more dimensions

tests/transforms/test_transformations.py
```diff
@@ -0,0 +1,177 @@
...
+            (
+                # maximum downsample ND array
+                # not defined
+                np.array([
+                    [1, 3], [4, 4], [6, 8], [2, 1], [3, 2],
+                    [5, 1234], [11, 12]]),
+                7, 2, 0, 'maximum',
+                np.array([[6, 8], [11, 12]])),
+            ])
+def test_downsample_exceptions(array, input_fps, output_fps, random_seed,
```
kschelonka, line 143: +1

segmentation_labeling_app/transforms/transform_pipeline.py
djkapner: after discussion: we'll leave this in for now. Debatable how useful it is.

kschelonka: It's just a nit. Fine to leave it in.

segmentation_labeling_app/transforms/transform_pipeline.py
djkapner: after discussion: we are deciding to not implement a pre-filter at this time. This TODO will remain.

segmentation_labeling_app/rois/rois.py
djkapner: removed support for traceless ROIs and deleted this TODO

PR #43, opened 2020-04-25
Add test for transform_pipeline

njmei:


segmentation_labeling_app/transforms/transform_pipeline.py
```diff
@@ -127,27 +135,29 @@ def run(self, db_conn: query_utils.DbConnection):
...
-            manifest['max-source-ref'] = max_proj_path
-            manifest['avg-source-ref'] = ave_proj_path
+            manifest['source-ref'] = str(outline_path)
+            manifest['roi-mask-source-ref'] = str(mask_path)
+            manifest['video-source-ref'] = str(sub_video_path)
+            manifest['max-source-ref'] = str(max_proj_path)
+            manifest['avg-source-ref'] = str(avg_proj_path)
             # manifest['trace-source-ref'] =
 
+            save_manifest(manifest, output_dir)
```
djkapner, line 159: I think this function is fine for now, as a placeholder, but:
In the [design doc](https://github.com/AllenInstitute/informatics_design_docs/tree/master/ophys/segmentation#create-manifest-and-sqlite-entry) we said we are going to make one ROI manifest entry per row in a sqlite file.
Things got spread out over the design doc/zenhub tasks, and PRs. But #37 shows a checklist indicating that we need sqlite or (nice-to-have) sql table to hold these manifests.
This is out-of-scope for your PR. But, it looks like you needed some outputs for validation, so I'm fine with keeping it, as you established a nice test structure for all this and the next PR can modify when getting sql output in place.

PR #53, opened 2020-04-29
Feature/upload tweaks

djkapner:
3 minor changes:
* removes an extra `/` in the s3 URIs when no prefix is provided: `<bucket>//<timestamp>/<key>` -> `<bucket>/<timestamp>/<key>`
* sets an argschema arg for list input to get rid of annoying future warning.
* adds logging info output so user knows something is happening.

Sample output:
```
$ python -m slapp.transfers.upload --input_json input2.json 
INFO:LabelDataUploader:Requesting 11 roi manifests from postgres
WARNING:LabelDataUploader:Requested 11, received 10. Missing ids: {999999}
INFO:LabelDataUploader:bucket destination is s3://dev.slapp.alleninstitute.org/20200429074103
INFO:botocore.credentials:Found credentials in shared credentials file: ~/.aws/credentials
INFO:LabelDataUploader:uploaded source data for 10 / 10 ROIs
INFO:LabelDataUploader:uploaded s3://dev.slapp.alleninstitute.org/20200429074103/manifest.jsonl

```

### DIFF
--- a/slapp/transfers/upload.py
+++ b/slapp/transfers/upload.py
@@ -37,15 +37,22 @@ class LabelDataUploader(argschema.ArgSchemaParser):
         self.timestamp = datetime.datetime.now().strftime('%Y%m%d%H%M%S')
 
         # get the specified per-ROI manifests
+        nrequested = len(self.args['roi_manifests_ids'])
+        self.logger.info(f"Requesting {nrequested} roi manifests from postgres")
+
         idstr = repr(self.args['roi_manifests_ids'])[1:-1]
-        query_string = ("SELECT manifest FROM roi_manifests "
+        query_string = ("SELECT id, manifest FROM roi_manifests "
                         f"WHERE id in ({idstr})")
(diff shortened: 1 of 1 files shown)


slapp/transfers/upload.py
```diff
@@ -38,26 +41,42 @@ def run(self, db_conn: query_utils.DbConnection):
         query_string = ("SELECT manifest FROM roi_manifests "
                         f"WHERE id in ({idstr})")
         manifests = [r['manifest'] for r in db_conn.query(query_string)]
+        nman = len(manifests)
+        self.logger.info(
```
isaak-willett, line 45: We might want to log a warning if they don't agree on length. I'm okay with it either way just a thought for a more verbose log message.

djkapner: Good idea, added that and updated the example output above.

isaak-willett: Looks good to me!

PR #54, opened 2020-04-29
deployment tweaks

djkapner:
### 2. linking movie frame rate and trace time axis
webapp movie playback defined by `fps` passed to imageio_ffmpeg and trace playback defined by `pointInterval`. These were not connected. Added a `playback_factor` arg default to 1.0 (real time) that connected these 2 data sources. Demo:
```
import json                                                               
tpath = "/allen/aibs/informatics/labeling_artifacts/segmentation_run_id_954/trace_1575130.json"                                                     
with open(tpath, "r") as f: 
    print(json.load(f)['pointInterval'])                                                                            
0.25
```
and
```
$ ffmpeg -i /allen/aibs/informatics/labeling_artifacts/segmentation_run_id_954/video_1575130.mp4
...
    Stream #0:0(und): Video: h264 (High) ... 4 fps, 4 tbr, 16384 tbn, 8 tbc (default)
...
```
note that in python:
```
import imageio_ffmpeg as mpg                                              
vpath = "/allen/aibs/informatics/labeling_artifacts/segmentation_run_id_954/video_1575130.mp4"                                                    
frames, seconds = mpg.count_frames_and_secs(vpath) 
print(frames / seconds)                                                  
4.047058823529412
```
`imageio-ffmpeg` has some [known issue](https://github.com/imageio/imageio-ffmpeg/blob/58524a3ac54352f945d2d62ecd938f3a2c1ac8a8/imageio_ffmpeg/_io.py#L23-L31).

### 3. Added a video/image normalization strategy.
Thus far the strategy is:
* downsample the video and take average and max projections.
* normalize video and average projection based on quantiles of average projection.
* normalize max projection based on quantiles of max projection. (saturates when treated like video and avg)
* normalization completely at experiment level.
* no per-ROI normalization (in this iteration.)

### 5. Added timestamp to local transform results folder.
To avoid re-writing some manifest contents. Could get confusing.

### DIFF
--- a/tests/transforms/test_transform_pipeline.py
+++ b/tests/transforms/test_transform_pipeline.py
@@ -137,6 +137,12 @@ def test_transform_pipeline(tmp_path, monkeypatch, mock_db_conn_fixture,
         mock_imageio.imsave.assert_has_calls(calls, any_order=False)
 
     # Assert that created manifests are correct
-    outdir_name = f"seg_run_id_{input_data['segmentation_run_id']}"
-    expected_outdir = Path(input_data["artifact_basedir"]) / outdir_name
-    [call(m, expected_outdir) for m in expected_manifests]
+    expected_insert_statements = [
+            transform_pipeline.insert_str_template.format(
+                manifest,
+                os.environ['TRANSFORM_HASH'],
+                manifest['roi-id'])
+            for manifest in expected_manifests]
+
+    mock_db_conn_fixture.bulk_insert.assert_has_calls(
+            [call(expected_insert_statements)])
--- a/tests/transforms/test_transform_pipeline.py
+++ b/tests/transforms/test_transform_pipeline.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from functools import partial
 import os
 import numpy as np
+import json
 
 from slapp.transforms import transform_pipeline
 
@@ -139,7 +140,7 @@ def test_transform_pipeline(tmp_path, monkeypatch, mock_db_conn_fixture,
     # Assert that created manifests are correct
     expected_insert_statements = [
             transform_pipeline.insert_str_template.format(
-                manifest,
+                json.dumps(manifest),
                 os.environ['TRANSFORM_HASH'],
                 manifest['roi-id'])
             for manifest in expected_manifests]
--- a/tests/transforms/test_transform_pipeline.py
+++ b/tests/transforms/test_transform_pipeline.py
@@ -90,17 +90,23 @@ def test_transform_pipeline(tmp_path, monkeypatch, mock_db_conn_fixture,
         manifest = create_expected_manifest(**meta, save_path=str(tmp_path))
         expected_manifests.append(manifest)
 
+    def normalize_side_effect(value, dummya, dummyb):
+        return value
+
     mock_downsample_h5_video = MagicMock(return_value=np.zeros((3, 5, 5)))
+    mock_normalize = MagicMock(side_effect=normalize_side_effect)
     mock_imageio = MagicMock()
     mock_content_extents = MagicMock(return_value=([0, 0, 2, 2], [1, 1, 1, 1]))
     mock_transform_to_mp4 = MagicMock()
     mock_np = MagicMock()
+    mock_np.quantile = MagicMock(return_value=(0, 1))
 
     mpatcher = partial(monkeypatch.setattr, target=transform_pipeline)
     mpatcher(name="ROI", value=mock_roi)
     mpatcher(name="downsample_h5_video", value=mock_downsample_h5_video)
     mpatcher(name="imageio", value=mock_imageio)
     mpatcher(name="content_extents", value=mock_content_extents)
+    mpatcher(name="normalize_array", value=mock_normalize)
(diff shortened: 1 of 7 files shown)


kschelonka:
Ensure traces and videos are on the same time domain. Currently videos are 10FPS and traces are 4 points per second.

djkapner:
TODO: add a test for normalize_array

tests/transforms/test_transform_pipeline.py
```diff
@@ -130,6 +137,6 @@ def test_transform_pipeline(tmp_path, monkeypatch, mock_db_conn_fixture,
         mock_imageio.imsave.assert_has_calls(calls, any_order=False)
 
     # Assert that created manifests are correct
-    outdir_name = f"segmentation_run_id_{input_data['segmentation_run_id']}"
+    outdir_name = f"seg_run_id_{input_data['segmentation_run_id']}"
     expected_outdir = Path(input_data["artifact_basedir"]) / outdir_name
-    expected_calls = [call(m, expected_outdir) for m in expected_manifests]
+    [call(m, expected_outdir) for m in expected_manifests]
```
njmei, line 142: Why was `expected_calls = [call(m, expected_outdir) for m in expected_manifests]` changed to `[call(m, expected_outdir) for m in expected_manifests]`? The line now doesn't do anything?

tests/transforms/test_video_utils.py
```diff
@@ -60,7 +61,10 @@ def test_mp4_conversion(video_fixture, tmp_path):
 
     fps = 30
 
-    transformations.transform_to_mp4(video=video_fixture,
+    normalized_video = normalize_array(
+            video_fixture, 0, video_fixture.max())
```
njmei, line 65: I would create a separate test called `test_mp4_conversion_with_normalized_video`. Since the current test if it failed, you wouldn't be able to **easily** tell of `normalize_array` failed or `transformations.transform_to_mp4` failed.

tests/transforms/test_transform_pipeline.py
djkapner: flake8 did not like that the variable `expected_calls` was assigned to and never used.
The line still executes `call()`:
```
In [29]: def myfun(i): 
    ...:     print(i) 
    ...:                                                                           

In [30]: [myfun(i) for i in range(4)]                                              
0
1
2
3
```

njmei: That `call()` is a helper for python mocking (https://docs.python.org/3/library/unittest.mock.html#call). So this line is saying setting up the expectation that something will have been called with the `expected_manifests` and the `expected_outdir`. There is a line missing then where `assert_has_calls` for a magic mock object is called.

njmei: It looks like what is missing is:

`mock_db_conn_fixture.bulk_insert.assert_has_calls(expected_calls)`

djkapner: odd, you did have this, and it looks like I removed it here: https://github.com/AllenInstitute/segmentation-labeling-app/pull/37/commits/86cfe430a27158615451f688f0ed77e24d99a25a as part of https://github.com/AllenInstitute/segmentation-labeling-app/pull/37
I'm not sure why, I will restore

njmei: Probably because I was literally saving a json output manifest there which has been replaced with saving to a database.

djkapner: oh, of course. thanks for refreshing my memory

tests/transforms/test_video_utils.py
djkapner: looking more closely, this test does not need to normalize the video, as I had removed video normalization from `transform_to_mp4` but, now that I am looking at it more closely, I think this test needs a little work.

djkapner: reworked this test in https://github.com/AllenInstitute/segmentation-labeling-app/pull/54/commits/240d1df4e08a2a79c8c6dc13ee0b94d588754c4a

tests/transforms/test_transform_pipeline.py
djkapner: added an `assert_has_calls` on `bulk_insert`

tests/transforms/test_video_utils.py
```diff
@@ -55,27 +39,33 @@ def test_video_downsample(
...
-    # this error unfortunately, this has been attempted and small errors
-    # persist
+    read_video = np.zeros((nframes_read, *meta['size']), dtype='uint8')
+    for i in range(nframes_read):
+        read_video[i] = reader.get_data(i)[:, :, 0]
+    reader.close()
 
-    assert frames.shape == video_fixture.shape
+    assert read_video.shape == video.shape
+    # mp4 is not lossless compression, so can't test for exact match
```
njmei, line 70: This is just a note

mp4 is merely the content container, the x264 codec that does the compression 'can' be lossless but you need to set an appropriate CRF:

https://slhck.info/video/2017/02/24/crf-guide.html

djkapner: changed comment to be more accurate

PR #84, opened 2020-05-09
Feature/dataset selection

djkapner:
Adds a module that logs dataset selection to a posttgres table, which is described in this [design doc PR](https://github.com/AllenInstitute/informatics_design_docs/pull/7).
Example showing (a little) repreoducible random selection:
image

### DIFF
--- a/slapp/data_selection/select_data.py
+++ b/slapp/data_selection/select_data.py
@@ -8,9 +8,9 @@ import numpy as np
 
 insert_statement_template = (
         "INSERT INTO experiment_selection "
-        "(base64_query_strings, slapp_commit_hash, "
-        "ophys_experiment_ids, sub_selected_ids, random_seed) "
-        "VALUES ({}, {}, {}, {}, {})")
+        "(base64_query_strings, slapp_commit_hash, ophys_experiment_ids, "
+        "sub_selected_ids, random_seed, comment_string) "
+        "VALUES ({}, {}, {}, {}, {}, {})")
 
 
 class DataSelectorSchema(argschema.ArgSchema):
@@ -30,6 +30,10 @@ class DataSelectorSchema(argschema.ArgSchema):
         required=False,
         default=42,
         description="random number generator seed")
+    comment_string = argschema.fields.Str(
+        required=False,
+        default="",
+        description="logged in postgres entry as comment_string")
 
     @mm.post_load
     def check_lengths(self, data, **kwargs):
@@ -67,7 +71,8 @@ class DataSelector(argschema.ArgSchemaParser):
             f"'{os.environ['TRANSFORM_HASH']}'",
             f"ARRAY{repr(experiment_ids)}",
             f"ARRAY{repr(sub_experiments)}",
-            f"{self.args['random_seed']}")
+            f"{self.args['random_seed']}",
+            f"'{self.args['comment_string']}'")
 
         label_dbconn.insert(insert_statement)
 
--- a/slapp/data_selection/select_data.py
+++ b/slapp/data_selection/select_data.py
@@ -50,11 +50,15 @@ class DataSelector(argschema.ArgSchemaParser):
     default_schema = DataSelectorSchema
 
     def run(self, lims_dbconn, label_dbconn):
+        self.logger.name = type(self).__name__
+
         experiment_ids = []
         b64_queries = []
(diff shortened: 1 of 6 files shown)


kschelonka:
Do we need to keep an eye on the table size since we're storing all the experiment_ids for every query? I'm not sure what our limitations are with the current dev DB setup.

slapp/data_selection/select_data.py
```diff
@@ -0,0 +1,113 @@
...
+            experiment_ids.append(
+                    [i['exp_id'] for i in lims_dbconn.query(qstring)])
+            self.logger.info(
+                    f"{qstring}\n returned {len(experiment_ids[-1])} ids")
+            b64_queries.append(
+                base64.b64encode(bytes(qstring, 'utf-8')).decode('utf-8'))
+
+        rng = np.random.default_rng(self.args['random_seed'])
+
+        sub_experiments = []
```
kschelonka, line 67: Why not [numpy.random.choice](https://numpy.org/doc/stable/reference/random/generated/numpy.random.choice.html?highlight=random%20choice#numpy.random.choice)?

slapp/utils/query_utils.py
```diff
@@ -17,22 +43,27 @@ def get_labeling_db_credentials() -> dict:
 
     Raises
     ------
-    LabelingEnvException
-        Raised if LABELING_USER/LABELING_PASSWORD are not set.
+    CredentialsException
+        Raised if user and password are not set as ENV variables
```
kschelonka, line 47: Typically we use all caps for env variables. I think "user" and "password" are a little generic, so I'd prefer the prior version (LABELING_USER, LABELING_PASSWORD).

slapp/data_selection/select_data.py
djkapner: You're right, instead of a shuffle and a selection, I can use choice with `replace=False`.
A slightly related point: numpy is moving away from `np.random.choice()` to 
```
rng = np.random.default_rng()
rng.choice()
```
https://numpy.org/doc/stable/reference/random/new-or-different.html#new-or-different

slapp/utils/query_utils.py
djkapner: That prior behavior is still in place and that structure of the variable names is still there. I just generalized this function with `env_prefix` so we can target different postgres servers. For example, now I have `LABELING_USER` and `LIMS_USER`. You can see these [here](https://github.com/AllenInstitute/segmentation-labeling-app/blob/bcdead70c5629651c1c1b212d47e1c855c0b86d2/slapp/data_selection/select_data.py#L88-L95)

djkapner:
> Do we need to keep an eye on the table size since we're storing all the experiment_ids for every query? I'm not sure what our limitations are with the current dev DB setup.

Generally, yes we shouldn't generate huge tables. But, I am not expecting more than 10-100 entries in this table.

slapp/data_selection/select_data.py
djkapner: switched to `choice` from `shuffle`

slapp/utils/query_utils.py
kschelonka: oh, I see, I misinterpreted. I find the documentation a little confusing though. The part about the list of keys is not in the section where you describe the format of the env variables. The code does uppercase the variables, but they're actually supposed to be all-caps when set in the env variable, so I think it's less confusing to document them as capitals (especially since they're not user-set in the function).

I would update it to be something like this:

```
    """Get labeling DB credentials from environment variables.
Environment variables must be in the format `{env_prefix}_{key}`,
where `key` is one of the following (case-sensitive): ['USER', 'PASSWORD', 'HOST', 'PORT', 'DATABASE']. Examples: LABELING_USER, LABELING_PASSWORD.
If values for `host`, `port`, and `database` are not passed as function arguments, then they must also be
set as environment variables using the `{env_prefix}_{key}` format, e.g. LABELING_HOST.

    Parameters
    ----------
    env_prefix : str
        expected environment variable prefix for credential keys.
        expected ENV variables are <prefix><KEY>. Values for <KEY> include the following: ["HOST", "PORT", "DATABASE", "USER", "PASSWORD"]. 
```

djkapner: Updated the docstring.

PR #86, opened 2020-05-11
Feature/full size movie

djkapner:
Adds:
- [x] updates [design doc](https://github.com/AllenInstitute/informatics_design_docs/pull/8)
- [x] save full-size downsampled movie to disk during transformation.
- [x] adds path to that movie to each `manifest` in `roi_manifests` entries.
- [x] uploads this object to S3 once per experiment. ROI manifests on S3 refer to shared full movie objects.
- [x] adds full-frame outline per ROI manifest.

screenshot showing S3 bucket contents, ordered by modified date. `full_movie` first, includes `full_outline` next ROI objects follow, `full_movie` not re-uploaded:
image

For this example, one frame of the movie with the generated `full_outline*.png` overlaid:
image

With these changes, I produced a truncated test movie with webm and uploaded to S3, then ran the UI from the upcoming full FOV UI PR https://github.com/AllenInstitute/segmentation-labeling-app/pull/52:

Peek 2020-05-13 12-17


slapp/transfers/upload.py
```diff
@@ -66,13 +72,29 @@ def run(self, db_conn: query_utils.DbConnection):
         uri = utils.s3_uri(self.args['s3_bucket_name'], prefix)
         self.logger.info(f"bucket destination is {uri}")
 
+        # upload the per-experiment object
+        full_video_path = np.unique([m['full-video-source-ref']
+                                     for m in manifests])
+        if len(full_video_path) != 1:
+            raise LabelDataException(
+                    "manifests do not share a single movie path")
+        object_key = prefix + "/" + pathlib.PurePath(full_video_path[0]).name
```
isaak-willett, line 81: I think we should write this using Path objects, makes it more portable.
object_key = `pathlib.PurePath(prefix) / pathlib.PurePath(full_video_path[0]).name`

slapp/transforms/transform_pipeline.py
```diff
@@ -170,25 +171,38 @@ def run(self, db_conn: query_utils.DbConnection):
...
             outline_path = output_dir / f"outline_{roi.roi_id}.png"
+            full_outline_path = output_dir / f"full_outline_{roi.roi_id}.png"
             sub_video_path = output_dir / f"video_{roi.roi_id}.webm"
             max_proj_path = output_dir / f"max_{roi.roi_id}.png"
             avg_proj_path = output_dir / f"avg_{roi.roi_id}.png"
             trace_path = output_dir / f"trace_{roi.roi_id}.json"
 
             mask = roi.generate_ROI_mask(
                     shape=self.args['cropped_shape'])
+            mask = np.uint8(mask * 255 / mask.max())
```
isaak-willett, line 194: I think we should normalize against the global roi_mask max or video max, so that from one roi to another there is a difference in intensity in the mask. ie: An roi with a max probability of 0.95 and one with a max probability of 0.7 will both end up having those pixels at 255, which I think is less informative to the viewer. I think we lose some knowledge in that transform.

djkapner: These are not probabilities, they are weights. I made this choice intentionally to make sure that the masks were visible on the scale. These users have not usually seen weighted masks, so I don't expect them to be able to parse the weight value and have that contribute to their judgment on the ROI. The mask weights in the image are valuable if they see that the low-weighted (darker areas) correspond with some pixels that don't appear to be part of the cell.

djkapner: I'd also mention that the mask was already being normalized before this change by `imageio.imsave()` but it was throwing an annoying warning that `you can get rid of this warning by doing the uint8 conversion yourself`, so I just added this to suppress that output.

Issue #95, opened 2020-05-12
investigate and implement speed-up for webm creation.


djkapner:
Our webm movie creation is 20x slower than the mp4 creation. (per ROI 300 seconds vs 15)

- [ ] ~ffmpeg has [multi-threading available](https://trac.ffmpeg.org/wiki/Encode/VP9#speed) for VP9. Determine how to call this functionality through imageio-ffmpeg and confirm that we can increase speed of writing.~

See notes below. Above point was not much more than 2x improvement in speed.

Implement time-chunked multiprocessing. Some headway [here](https://github.com/AllenInstitute/segmentation-labeling-app/blob/feature/ffmpeg_threading/slapp/transforms/video_utils.py). This needs:
- [x] fixed tests
- [x] docstring for `transform_job`
- [x] `transform_pipeline` arg for controlling ncpu. (default to -1 should set to `multiprocessing.get_cpu_count()`)
- [x] general review.

djkapner:
this was looking more like a must-have, given how slow things were. I experimented with `row-mt`, passed into our imageio-ffmpeg writer as:
```
 writer = mpg.write_frames(output_path,
                              (video[0].shape[1], video[0].shape[0]),
                              video[0].shape[::-1],
                              pix_fmt_in="gray8",
                              pix_fmt_out="yuv420p",
                              codec="libvpx-vp9",
                              fps=fps,
                              bitrate=bitrate)
                              bitrate=bitrate,
                              output_params=['-row-mt', '1'])
```
but that only gave a 30-40% speeedup. Not close to recovering the 20x.
A chunk-by-time multiprocessing approach with an ffmpeg concat seems to be working better and

Issue #94, opened 2020-05-12
Fix blurry video artifacts


isaak-willett:
# Overview
![Screen Shot 2020-05-12 at 1.50.22 PM.png](https://images.zenhubusercontent.com/5dde9d05537fdc0001b7a121/bf7d2f7d-d4f2-4694-924a-466aae49de7d)
It can be seen from the above screen shot that the objects in the frame are blurry. This could be the result of codec issues using constant bit rate over constant quality. To maintain constant quality in a video the bitrate must change throughout, restricting the bit rate leaves highly changing parts of the video without the quality of less changing sections.

Initial options to check out:
- [ ] constant quality encoding in imageio_ffmpeg
- [ ] Non constant bit rate change to provide less blurry artifacts in videos

# Notes
[ffmpeg constant quality docs](https://trac.ffmpeg.org/wiki/Encode/VP9)
[bit rates in videos](https://help.encoding.com/knowledge-base/article/understanding-bitrates-in-video-files/)
[ffmpeg commands in imageio](https://imageio.readthedocs.io/en/stable/format_ffmpeg.html)

# Validation Criteria:
Both full-size and 128x128 videos play on modern Chrome and Firefox browsers. The videos load and play consistently, line up with the traces, there are no visible artifacts, and they can be used by labelers to make classification decisions. This should be demonstrated for videos that come from at least five experiments, since artifacts are not present for every video. For testing, create a simple webpage that plays the videos from local files. For final validation, create a SageMaker job and invite the dev team to test.

njmei:
imageio-ffmpeg looks to literally just stitch together an ffmpeg command. Here is the relevant line for what modifying 'bitrate' does. Which is probably not quite the desired thing for the vp9 encoder.

https://github.com/imageio/imageio-ffmpeg/blob/974e92e699676822971d66765a2787ecac3a0788/imageio_ffmpeg/_io.py#L383

kschelonka:
Updated title and validation criteria.

njmei:
Resolved by #100

PR #100, opened 2020-05-14
Parallelize video encoding and add quality (crf) control for encoding

njmei:
This commit aims to solve the problem of slow VP9 encoding by
implementing parallelization of the webm transform step. Videos
are split into chunks by time and separately encoded via
multiprocessing before being stitched back together with ffmpeg.

This commit also adds the ability to specify a desired perceptual
quality for the encoded video.

The default encoding strategy has been changed from
"constant bitrate" to "constant quality". For more details please
see: https://trac.ffmpeg.org/wiki/Encode/VP9

Relates to: #94, #95

------
Still TODO:

Run updated pipeline with following `ophys_experiment_ids`:
- [x] 850517352
- [x] 850517344
- [x] 853988430
- [x] 856123119
- [x] 854759894

- [x] If ophys_experiment_id videos look okay, try spinning up entire labeling app

- [x] Squash commits when everything looks good

### DIFF
--- a/slapp/transforms/transform_pipeline.py
+++ b/slapp/transforms/transform_pipeline.py
@@ -1,18 +1,19 @@
-import argschema
-import marshmallow as mm
-from pathlib import Path
-import os
-import json
-import imageio
-import numpy as np
 import datetime
+import json
+import multiprocessing
+import os
+from pathlib import Path
+
+import argschema
+import imageio
+import marshmallow as mm
+import numpy as np
+
 import slapp.utils.query_utils as query_utils
 from slapp.rois import ROI
-from slapp.transforms.video_utils import (
-    downsample_h5_video, transform_to_webm)
-from slapp.transforms.array_utils import (
-        content_extents, downsample_array, normalize_array)
-
+from slapp.transforms.array_utils import (content_extents, downsample_array,
+                                          normalize_array)
+from slapp.transforms.video_utils import downsample_h5_video, transform_to_webm
 
 insert_str_template = (
         "INSERT INTO roi_manifests "
@@ -86,8 +87,22 @@ class TransformPipelineSchema(argschema.ArgSchema):
                      "histogram adjustment"))
     webm_bitrate = argschema.fields.Str(
         required=False,
-        default="192k",
+        default="0",
         description="passed as bitrate to imageio-ffmpeg.write_frames()")
+    webm_quality = argschema.fields.Int(
+        required=False,
+        default=30,
+        description=("Governs encoded video perceptual quality. "
+                     "Can be from 0-63. Lower values mean higher quality. "
+                     "Passed as crf to ffmpeg")
+    )
+    webm_parallelization = argschema.fields.Int(
+        required=False,
+        default=1,
+        description=("Number of parallel processes to use for video encoding. "
+                     "A value of -1 results in "
+                     "using multiprocessing.cpu_count()")
+    )
 
     @mm.pre_load
     def set_segmentation_run_id(self, data, **kwargs):
@@ -117,6 +132,12 @@ class TransformPipelineSchema(argschema.ArgSchema):
                 raise TransformPipelineException(
                     f"{query_string} did not return exactly 1 result")
             data['segmentation_run_id'] = entries[0]['id']
+        return data
+
+    @mm.post_load
+    def set_webm_parallelization(self, data, **kwargs):
+        if data["webm_parallelization"] == -1:
+            data["webm_parallelization"] = multiprocessing.cpu_count()
         return data
 
 
@@ -174,8 +195,9 @@ class TransformPipeline(argschema.ArgSchemaParser):
         # experiment-level artifact
         full_video_path = output_dir / "full_video.webm"
         transform_to_webm(
-                downsampled_video, str(full_video_path),
-                playback_fps, self.args['webm_bitrate'])
+            video=downsampled_video, output_path=str(full_video_path),
+            fps=playback_fps, ncpu=self.args['webm_parallelization'],
+            bitrate=self.args['webm_bitrate'], crf=self.args['webm_quality'])
 
         # create the per-ROI artifacts
         insert_statements = []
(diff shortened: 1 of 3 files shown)


slapp/transforms/video_utils.py
```diff
@@ -46,32 +51,130 @@ def downsample_h5_video(
...
-                              (video[0].shape[1], video[0].shape[0]),
+                              video_shape,
                               pix_fmt_in="gray8",
                               pix_fmt_out="yuv420p",
                               codec="libvpx-vp9",
                               fps=fps,
-                              bitrate=bitrate)
-    writer.send(None)
+                              bitrate=bitrate,
+                              output_params=["-crf", str(crf)])
```
njmei, line 125: Note: `crf` has to be set via `output_params` because `imageio-ffmpeg` does the wrong thing if you try to set `quality`:

https://github.com/imageio/imageio-ffmpeg/blob/master/imageio_ffmpeg/_io.py#L385-396

slapp/transforms/transform_pipeline.py
```diff
@@ -86,8 +87,20 @@ class TransformPipelineSchema(argschema.ArgSchema):
...
+        default="0",
         description="passed as bitrate to imageio-ffmpeg.write_frames()")
+    webm_quality = argschema.fields.Int(
+        required=False,
+        default=20,
+        description=("Governs encoded video perceptual quality. "
+                     "Can be from 0-63. Lower values mean higher quality. "
+                     "Passed as crf to ffmpeg")
+    )
+    webm_parallelization = argschema.fields.Int(
```
kschelonka, line 99: Nit-ish: The default here should probably be 1. If we're running it on the cluster, you could be asking for more resources than are allocated.

slapp/transforms/transform_pipeline.py
```diff
@@ -86,8 +87,20 @@ class TransformPipelineSchema(argschema.ArgSchema):
...
+        required=False,
+        default=multiprocessing.cpu_count(),
```
kschelonka, line 101: Nit-ish: The default here should probably be 1. If we're running it on the cluster, you could be asking for more resources than are allocated.

tests/transforms/test_video_utils.py
```diff
@@ -39,39 +42,136 @@ def test_video_downsample(
...
+    for frame in reader:
+        parsed_frame = np.frombuffer(frame, dtype='uint8')
+        parsed_frame = parsed_frame.reshape(meta["size"][::-1])
+        obt_frames.append(parsed_frame)
+    obt_video = np.array(obt_frames)
+
+    assert obt_video.shape == expected_video.shape
+    # the default settings for imageio-ffmpeg are not lossless
+    # so can't test for exact match
+    np.testing.assert_allclose(obt_video, expected_video, atol=20)
```
kschelonka, line 66: 👍

tests/transforms/test_video_utils.py
```diff
@@ -39,39 +42,136 @@ def test_video_downsample(
     assert np.array_equal(downsampled_video, expected)
 
 
-@pytest.mark.parametrize("video_size", [((16, 32)),
-                                        ((16, 16))])
-def test_webm_conversion(video_size, tmp_path):
+def cmp_videos(encoded_video_path: str, expected_video: np.ndarray):
```
kschelonka, line 45: Nit: `compare` vs `cmp` is much more readable

kschelonka:
In general LGTM from reading tests. I'd like to see the outputs though.

slapp/transforms/transform_pipeline.py
njmei: I've changed this to 1 by default now.

tests/transforms/test_video_utils.py
njmei: Changed `cmp_videos` to `compare_videos`.

njmei:
All `crf` set to `30`. Videos look decent to me but having additional eyes would be great!

/allen/scratch/aibstemp/nicholasm/slapp_xform_pipeline_output

`segmentation_run_id 966` `8.70 minutes` `full_video_size=225.8MB`
`segmentation_run_id 967` `12.6 minutes` `full_video_size=595.6MB`
`segmentation_run_id 973` `14.8 minutes` `full_video_size=737.2MB`
`segmentation_run_id 972` `13.2 minutes` `full_video_size=752.7MB`
`segmentation_run_id 970` `8.95 minutes` `full_video_size=619.9MB`

slapp/transforms/video_utils.py
```diff
@@ -46,32 +51,130 @@ def downsample_h5_video(
...
+
+    Parameters
+    ----------
+    video : np.ndarray
+        Video to be transformed with shape (time, row, col)
+    output_path : str
+        Output path for the transformed video
+    fps : float
+        Desired frames per second (fps) of the output video
+    ncpu : int
```
djkapner, line 149: I think there should be a mechanism for setting this to [the number of cpus on the system](https://docs.python.org/3.4/library/multiprocessing.html?highlight=process#multiprocessing.cpu_count). That way, if we are running jobs with `#PBS -n` it will utilize all the CPUs on a node whether it lands on a 24, 32, or 40.
I think a value of `-1` could suffice to trigger that behavior.

njmei: Unclear whether the number of python `multiprocessing` module is smart enough to realize that not all CPUs are available. I think it's safer to require specifying `ncpu` in the actual qsub script where you can specify `#PBS -l ncpus=#`.

slapp/transforms/video_utils.py
```diff
@@ -46,32 +51,130 @@ def downsample_h5_video(
...
+    -------
+    str
+        Output path of the encoded video
     """
 
-    # ffmpeg expects the video shape in width, height not row, col
-    # have to reverse shape when inputting
-    # gray8 is uint8 format
+    # ffmpeg expects video shape in terms of: (width, height)
+    video_shape = (video[0].shape[1], video[0].shape[0])
```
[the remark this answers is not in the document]
njmei: I separated this out on purpose to be explicit about how imageio (ffmpeg) expects the shape to be passed in.

slapp/transforms/video_utils.py
kschelonka: @djkapner Yeah this was by my request, since if we run on the cluster not all on the node are necessarily available.

slapp/transforms/video_utils.py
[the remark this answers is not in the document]
djkapner: i don't, you can leave it.

PR #102, opened 2020-05-15
Adds a scale to the bottom of images and videos

isaak-willett:
# Overview

This commit adds a scale onto the bottom of images and videos so that
users have a better idea of the size of objects in view for more informed
labeling. Added a new module, image_utils.py, to hold the add_scale function.
Add scale_function adds a scale of user specified size in microns
to the bottom of a supplied image. Transformation pipeline was updated
to make use of this function to add scales to outline masks, in full and
sub-sampled views. The testing functions were updated to work with this
change.

# Adds
- [x] Scale function in new module image_utils
- [x] Changes to transformation pipeline to add the scale to mask outlines
- [x] Testing function for scale function
- [x] Changes to transformation pipeline test to keep the test working

# Validation

### Full ROI Mask Outline Image with Scale
full_outline_1575089

### Subset ROI Mask Outline Image with Scale
outline_1575089

### Sagemaker GT Screenshots
image
demo

### Sagemaker GT Job Link
[Ground Truth Job](https://nam12.safelinks.protection.outlook.com/?url=https%3A%2F%2Fuk2ac5ikj9.labeling.us-west-2.sagemaker.aws%2F&data=02%7C01%7C%7C92306f014b804907403a08d7ed4f965d%7C32669cd6737f4b398bddd6951120d3fc%7C0%7C1%7C637238798395243851&sdata=RutdM%2FnhliTK4i8vIM271mBNrvUP80GJTC5n4uIfwGM%3D&reserved=0)


slapp/transforms/image_utils.py
```diff
@@ -0,0 +1,63 @@
+from typing import Tuple
+from pathlib import Path
+from PIL import Image, ImageDraw, ImageFont
+from math import ceil
+
+
+def add_scale(image_path: Path,
+              scale_position: Tuple[int, int],
+              font_file: Path,
+              ophys_experiment_resolution: float = 10/13,
```
kschelonka, line 10: This is a generally applicable function, so let's call this `resolution`.

slapp/transforms/image_utils.py
```diff
@@ -0,0 +1,63 @@
...
+              scale_size: int = 10,
+              fill: Tuple[int, int, int] = (0, 0, 0),
```
kschelonka, line 12: Nit: `fill_color` would make this clearer without having to read the docstring.

slapp/transforms/image_utils.py
```diff
@@ -0,0 +1,63 @@
...
+              width: int = 5,
```
kschelonka, line 13: Nit-ish: `line_weight` ("width" to me makes it seem like the width the line spans, which is determined by the resolution and scale_size).

slapp/transforms/image_utils.py
```diff
@@ -0,0 +1,63 @@
...
+    font_file: Path
+        the font file to be used as denoted by a path object. On windows can
+        just be specified as 'arial.ttf' on linux must be a full path.
+    ophys_experiment_resolution: float = 10/13
+        resolution of the video in microns per pixel, defaulted to
+        scientifica general which is 10/13~.78 micros/pixel
+    scale_size: int=10
+        how many micro meters the scale bar should display, defaulted to 10
+    fill: Tuple=(0, 0, 0)
+        the fill color to draw the text and scale, defaults to black
```
kschelonka, line 34: Note rgb tuple

slapp/transforms/image_utils.py
```diff
@@ -0,0 +1,63 @@
+from typing import Tuple
+from pathlib import Path
+from PIL import Image, ImageDraw, ImageFont
+from math import ceil
+
+
+def add_scale(image_path: Path,
```
kschelonka, line 7: Would be nice to overload this and pass either a path or the image data

isaak-willett:
Just an FYI I'm going to rebase away a couple of commits before addressing changes to clean up history. I'm also going to improve the PR overview, then circle back around to comments.

kschelonka:
LGTM overall, I think it would be nice to see some images in the PR description.

slapp/transforms/image_utils.py
isaak-willett: changed name to resolution

slapp/transforms/image_utils.py
isaak-willett: changed to fill_color

slapp/transforms/image_utils.py
isaak-willett: changed to line_weight

slapp/transforms/image_utils.py
isaak-willett: noted rgb tuple

slapp/transforms/image_utils.py
isaak-willett: It's possible to do but it will be somewhat ugly, I tried this and you can make a PIL Image object from a numpy array but you can't draw or write to it. So I would have to make a temp dir to save the numpy array as a png to reload it into PIL as a png. The functionality to go from numpy -> Image in PIL (that has all features) is not there unfortunately.

kschelonka:
I'll approve after fixing linting errors. For the full pipeline I think we should move the scale up and to the right a few pixels. It looks too close to the edge IMO on the images you showed.

slapp/transforms/image_utils.py
kschelonka: But a PIL image should be fine, right?

slapp/transforms/transform_pipeline.py
```diff
@@ -88,6 +90,26 @@ class TransformPipelineSchema(argschema.ArgSchema):
         required=False,
         default="192k",
         description="passed as bitrate to imageio-ffmpeg.write_frames()")
+    font_file = argschema.fields.Str(
```
djkapner, line 93: you can change `Str` to `InputFile` and then argschema will validate and raise exceptions for you and you will not need your `mm.pre_load` below.

slapp/transforms/transform_pipeline.py
```diff
@@ -88,6 +90,26 @@ class TransformPipelineSchema(argschema.ArgSchema):
...
+    def validate_font_file(self, data, **kwargs):
+        """
+        validate that the font file provided exists in the system, throw
+        an exception if it doesn't.
+        """
+        if 'font_file' in data:
+            if not os.path.exists(data['font_file']):
+                raise TransformPipelineException(
+                    "Provided font file path does not exist. Please "
+                    "enter a valid font file path."
```
djkapner, line 110: If you do end up keeping this `pre_load` and I don't  think you need to, you could tell the user what file does not exist. The argschema validation will explicitly tell you "this parameter was suppposed to be a readable file and /blah/blah/file.txt is not readable".

slapp/transforms/transform_pipeline.py
```diff
@@ -204,6 +226,22 @@ def run(self, db_conn: query_utils.DbConnection):
             imageio.imsave(outline_path, outline, transparency=0)
             imageio.imsave(full_outline_path, full_outline, transparency=0)
 
+            # add the scale to the outlines they will toggle with the outline
+            # element on the web app
+            scale_position = (0, outline.shape[1])
+            scale_position_full = (0, full_outline.shape[1])
+            font_file_path = Path(self.args['font_file'])
+
+            outline_with_scale = add_scale(outline_path, scale_position,
```
djkapner, line 235: I think it would be better if your `add_scale` function accepts an Image or numpy array rather than a path. Because now the transform pipeline writes to disk, reads immediately and then overwrites. It is not a huge overhead, but, it's best to just do things in memory.

slapp/transforms/image_utils.py
```diff
@@ -0,0 +1,71 @@
...
+        x, y format, this is the format PIL expects.
+    font_file: Path
+        the font file to be used as denoted by a path object. On windows can
+        just be specified as 'arial.ttf' on linux must be a full path.
+    resolution: float = 10/13
+        resolution of the video in microns per pixel, defaulted to
+        scientifica general which is 10/13~.78 micros/pixel
+    scale_size: int=10
+        how many micro meters the scale bar should display, defaulted to 10
+    fill_color: Tuple=(0, 0, 0)
```
djkapner, line 35: for `fill_color` and `line_weight` you are passing these right into `draw()`, I think it makes sense not to rename them to be exactly the same as the keywords for `draw()` and to note in your docstring that they are being directed to those keywords in draw(). Then you don't have to redefine them. Reader can refer to docs for `draw()`.

slapp/transforms/image_utils.py
```diff
@@ -0,0 +1,71 @@
...
+from math import ceil
+
+import numpy as np
+from PIL import Image, ImageDraw, ImageFont
+
+
+def add_scale(image: Union[Path],
+              scale_position: Tuple[int, int],
+              font_file: Path,
+              resolution: float = 10 / 13,
```
djkapner, line 12: this is not exact. If you're going to specify a float, why not make it exact. `400 / 512` or `25 / 32` or `0.78125`

slapp/transforms/image_utils.py
djkapner: I see this is in direct opposition to what @kschelonka asked for earlier. I disagree, but you don't have to act on this one.

kschelonka: @djkapner In general I agree with trying to keep the parameters. If it were explicitly mentioned in the docstring I think it would be fine. The main reason I asked for changes is because I think those names make sense for the line drawing function, but are more confusing in the sense of a scale drawing -- especially "width".

slapp/transforms/transform_pipeline.py
isaak-willett: Good suggestion, changed it to InputFile

slapp/transforms/transform_pipeline.py
isaak-willett: Removed the str and made input file!

slapp/transforms/transform_pipeline.py
isaak-willett: Changed to in memory, I had made a previous comment to Kat that this was difficult but I resolved that. I had forgotten to convert to RGBA before drawing in PIL which causes errors. Everything is now in memory and only written once.

slapp/transforms/image_utils.py
isaak-willett: I'll revert these and have the params reflect the PIL params.

slapp/transforms/image_utils.py
isaak-willett: Good call

slapp/transforms/image_utils.py
isaak-willett: It is I made a mistake and didn't convert the numpy array to RGBA when loading so drawing was failing.

slapp/transforms/image_utils.py
```diff
@@ -0,0 +1,75 @@
...
+              scale_size: int = 10,
+              fill: Tuple[int, int, int] = (0, 0, 0),
+              width: int = 5,
+              font_size: int = 8) -> Image:
+    """
+    Adds a scale bar onto a image passed in the form of a image path. Uses
+    PIL library ImageDraw and ImageFont to draw scale bar onto images.
+    Parameters
+    ----------
+    image: Path
```
djkapner, line 22: needs updating

kschelonka:
Please update your commit history prior to merging. The headers are not informative atm.

slapp/transforms/image_utils.py
isaak-willett: Updated the docstring

isaak-willett:
> 
> 
> Please update your commit history prior to merging. The headers are not informative atm. You can squash where it makes sense.

I rebased away the intermediate commits and left the one additional change Dan requested.

tests/transforms/test_transform_pipeline.py
```diff
@@ -132,15 +138,10 @@ def normalize_side_effect(value, dummya, dummyb):
...
-        outline_save = call(Path(manifest['source-ref']), ANY,
-                            transparency=0)
-        full_outline_save = call(
-                Path(manifest['full-outline-source-ref']), ANY, transparency=0)
         max_save = call(Path(manifest['max-source-ref']), ANY)
         avg_save = call(Path(manifest['avg-source-ref']), ANY)
 
-        calls = [mask_save, outline_save, full_outline_save,
-                 max_save, avg_save]
+        calls = [mask_save, max_save, avg_save]
```
djkapner, line 144: I see why you are not checking these calls any more and that you're actually writing to a `tmp_path` with `Image.save()`. In place of asserting these calls, I think you should at least check that the expected files exist and potentially also read them and assert that the size and datatypes are as expected.

isaak-willett: The images aren't ever created in the test pipeline which is why I didn't initially add an assertion. I added an assertion that the add_scale function is called and add_scale.save() is called for every outline that is created. This falls in line with how Nick previously built the test.

PR #107, opened 2020-05-19
writes scale with opencv, rewrites tests, saves all with imageio

djkapner:
The outline images coming out of the transform pipeline were showing as all red. Prior to adding the scale bars, we used a call such as `imageio.save(..., transparency=0)` which assigns the pixel value of 0 to be transparent. The outline values were stored as `1` because the front-end wanted black (technically `0` would be black, but close enough) to be the outline.
Another method for png to store transparency is with an alpha channel. One can start reading about these 2 transparency methods for png [here](https://en.wikipedia.org/wiki/Portable_Network_Graphics#Transparency_of_image).
The `add_scale` method using `PIL.Image` was attempting to use an alpha channel. I suspected that it was not configured properly, as the entire alpha channel was set to `255`.
I started to try to fix this using `PIL.Image` but I found that method was cumbersome. I really wanted to get back to using `imageio.imsave()` since I knew that worked previously and iterating through the AWS front-end is a little slow.

Some benefits of using `opencv`:
* I find the `add_scale()` method to be cleaner and more readable now.
* `opencv` methods will operate on arrays and easily return arrays.
* `opencv` has some fonts already packaged. No need to chase font files around.

Some downsides to using `opencv`:
* It is not the best font I've ever seen. I don't think Greek letters are possible.

Some changes:
* `add_scale()` is no longer overloaded. It takes an array, it gives back an array. We're not using it with `Path` input.
* refactored the tests for `add_scale()`. These aren't complete in that the text never makes it into the frame, and multi-pixel line widths aren't straightforward, as the line ends get rounded. I don't test these tough cases. I am just showing how to move the scale around.
* exposed the `add_scale` args to the pipeline user.

Manifest with 5 ROIs made with this branch: `s3://dev.slapp.alleninstitute.org/trial/seg_run_id_958/20200518214352/manifest.json`

Peek 2020-05-19 12-10


slapp/transforms/image_utils.py
```diff
@@ -1,76 +1,56 @@
...
-    font_size: Int
-        the font size of the description text
+        image coordinates for the corner of the scale bars
+    um_per_pixel: float = 400/512
+        resolution of the video in microns per pixel
+    scale_size_um: float
+        length in microns of scale bars
+    color: int
+        default 1. passed as `color` to cv2.line and cv2.putText
+    thickness_um: float
```
kschelonka, line 28: Nit: Why do the line thickness in microns? Pixels makes more sense to me since the line thickness has nothing to do with the resolution.

slapp/transforms/image_utils.py
```diff
@@ -1,76 +1,56 @@
...
-        name in that API. The width of the lines of the scale in pixels
-    font_size: Int
-        the font size of the description text
+        image coordinates for the corner of the scale bars
+    um_per_pixel: float = 400/512
+        resolution of the video in microns per pixel
+    scale_size_um: float
+        length in microns of scale bars
+    color: int
+        default 1. passed as `color` to cv2.line and cv2.putText
```
kschelonka, line 27: Would appreciate more explanation here, so I don't have to look up what color is in cv2. Is it a 0-1 scale, where 1 is black?

slapp/transforms/image_utils.py
djkapner: the only reason is that the `scale_size` is in um, which makes sense since we are annotating in um and providing `um_per_pixel`. With both `scale_size` and `thickness` in um, we can understand the aspect ratio of the line we are about to draw.
Your way makes sense also. I am happy to change it.

slapp/transforms/image_utils.py
```diff
@@ -1,76 +1,56 @@
...
+    thickness_um: float
+        converted to pixels and passed as `thickness` to
+        cv2.line and cv2.putText
+    fontScale: float
+        default 0.3. passed as `fontScale` to cv2.fontScale
```
kschelonka, line 32: More explanation here again would be nice. 

The base size of the font is multiplied by this scale factor. Passed as `fontScale` to cv2.fontScale

slapp/transforms/image_utils.py
djkapner: I'm not sure where this came about in our development, but, the front-end wants a black line on a transparent background. Black is `0` and white is `255`. We are setting `transparency=0` so we can't really draw in black. So I set it to 1 (or it was set to 1, can't remember).
The right way to fix this is for `generate_ROI_outline` to return a black (`0`) outline on a white (`255`) background and then tell `imageio` `transparency=255`. I'm happy to make that change.

slapp/transforms/transform_pipeline.py
```diff
@@ -113,6 +107,28 @@ class TransformPipelineSchema(argschema.ArgSchema):
...
+    scale_offset = argschema.fields.Int(
+        required=False,
+        default=3,
+        description=("number of pixels scale corner is offset from "
+                     "lower left"))
+    full_scale_offset = argschema.fields.Int(
+        required=False,
+        default=12,
+        description=("number of pixels scale corner is offset from "
+                     "lower left"))
```
kschelonka, line 119: "...in the full-field-of-view ROI outline."

slapp/transforms/transform_pipeline.py
```diff
@@ -113,6 +107,28 @@ class TransformPipelineSchema(argschema.ArgSchema):
                      "A value of -1 results in "
                      "using multiprocessing.cpu_count()")
     )
+    scale_offset = argschema.fields.Int(
+        required=False,
+        default=3,
+        description=("number of pixels scale corner is offset from "
+                     "lower left"))
```
kschelonka, line 114: "...in the ROI outline"

kschelonka: Repeat for scale size

slapp/transforms/transform_pipeline.py
```diff
@@ -113,6 +107,28 @@ class TransformPipelineSchema(argschema.ArgSchema):
...
+        default=10.0,
+        description="length of scale bars in um")
+    full_scale_size_um = argschema.fields.Float(
+        required=False,
+        default=40.0,
+        description="length of scale bars in um")
+    um_per_pixel = argschema.fields.Float(
+        required=False,
+        default=0.78125,
+        description="microns per pixel")
```
kschelonka, line 131: "... in the 2p source video"

slapp/transforms/image_utils.py
kschelonka: I can see both sides. I personally would prefer pixels because I'm not thinking about the aspect ratio as much as I am the actual thickness of the line in pixels on the screen. But that's why this is a nit. I'll leave it up to you.

slapp/transforms/image_utils.py
kschelonka: What you described would make more sense. 

The reason the front-end uses black is because it uses color rotation to produce different colors for colorblind mode vs. default colors, rotated from black (which is a fairly straightforward computation).

kschelonka:
Overall it's in a good place. I think the minor documentation updates will make things more obvious to the naive user. And inverting the transparency to 255 vs. 0 as discussed so that the outline is actually black as expected.

slapp/transforms/image_utils.py
djkapner: Will leave in um.

slapp/transforms/image_utils.py
djkapner: implemented

slapp/transforms/image_utils.py
djkapner: see `Notes` section in docstring

slapp/transforms/transform_pipeline.py
djkapner: implemented

slapp/transforms/transform_pipeline.py
djkapner: implemented

slapp/transforms/transform_pipeline.py
djkapner: implemented

djkapner:
updated gif in PR description after revision commits

slapp/transforms/image_utils.py
```diff
@@ -24,17 +24,45 @@ def add_scale(array: np.ndarray,
...
+    produces text that is 7 pixels high. opencv provides a utility
+    to check this:
+
+    >>> cv2.getTextSize('10um', cv2.FONT_HERSHEY_SIMPLEX, 0.3, 1)
+    ((28, 7), 3)
+
+    where the second element of the tuple is the height of the bounding
+    box of the text. These results depend not only on fontScale but also
+    thickness. For a line thickness of 1, some examples of (fontScale, height)
+    are (0.3, 7), (0.7, 16), (1.0, 22)
```
kschelonka, line 64: Nice example clarifying this behavior.

kschelonka:
The font is ugly, but everything else looks great. Thanks!

PR #126, opened 2020-09-09
adds codecov config to warn only

djkapner:
stops flagging PRs with red X if codecov is not satisfied. Still comments on PR.


kschelonka (approved):
🌵

PR #131, opened 2022-03-03
allow masks larger than crop size

aamster:
Currently, if a mask was larger than the crop size, then it would not be padded.
This PR first center crops the mask to the crop size so that it is always <= crop size. This will cut off part of the mask if it is very large.

## Before
before

## After
after

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,3 +17,4 @@ pandas
 argschema==2.0.1
 jsonlines
 flake8
+imgaug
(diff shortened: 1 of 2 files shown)


danielsf (approved):
Cropping was not a problem in the max and avg projection images for some reason. Have you confirmed that the cropped mask is centered in the same way as those thumbnails (so that the mask can consistently be applied to the projection images)?

Probably doesn't actually matter since, again, these are very obviously not cells.

aamster:
It deals with projections correctly (img size is (128, 128)) and it lines up with the mask

